Issue #12, opened 2018-12-07
DAL: Spec Database for MongoDB


shlomi-nx:
Implement https://github.com/nutanix/papiea-js/blob/master/src/databases/spec_db_interface.ts for MongoDB

shlomi-nx:
As stated in the PR (#29), we should adopt `async/await` model instead of callbacks

PR #29, opened 2018-12-13
WIP: Spec Database for MongoDB

adolgarev:


src/databases/spec_db_mongo.ts
```diff
@@ -0,0 +1,38 @@
...
+        return [true, entity_metadata, spec];
+    }
+
+    get_spec(entity_ref: core.Entity_Reference):[core.Metadata, core.Spec] {
+        let entity_metadata:core.Metadata = {
+            uuid: "uuid",
+            kind: "kind",
+            spec_version: 0,
+            created_at: new Date(),
+            delete_at: new Date()
```
shlomi-nx: This should need to be initiated to null. This will be used later to allow for automatic deletion of the entity at a certain time. We may need to express this better in its type.

shlomi-nx:
resolve #12

shlomi-nx:
I spent a good amount of time learning javascript models for asynchronous programming. Instead of the callback solution proposed here (which may lead to really unmanageable code), I'd rather us use `async/await` which is a syntax sugar for the `Promise` monad-like structure (reminds me of trampoline if you are familiar).

In any case, lets reformulate this as `async/await`. You can read more here:
https://javascript.info/async-await
https://dzone.com/articles/from-callbacks-to-async-await-a-migration-guide
https://blog.usebutton.com/demystifying-async-programming-in-javascript-47e417566f61

Please let me know your thoughts

PR #30, opened 2018-12-14
Excluded cljs files & build from code coverage, fixed bug with coverage not seeing some files

Emulebest:


adolgarev:
screenshot_20181217_191717

PR #34, opened 2018-12-19
DAL async/await implementation

### LINKED ISSUE #12: DAL: Spec Database for MongoDB

### LINKED ISSUE #13: DAL: Status Database for MongoDB
Implement https://github.com/nutanix/papiea-js/blob/master/src/databases/status_db_interface.ts on Mongo initially

### LINKED ISSUE #14: DAL: Provider Database for MongoDB
Implement https://github.com/nutanix/papiea-js/blob/master/src/databases/providers_db_interface.ts.

Emulebest:
Resolves #12 
Resolves #13 
Resolves #14


src/databases/spec_db_mongo.ts
```diff
@@ -0,0 +1,71 @@
...
+            if (result.result.n !== 1) {
+                throw new Error(`Amount of updated entries doesn't equal to 1: ${result.result.n}`)
+            }
+            entity_metadata.spec_version++;
+            return [entity_metadata, spec]
+        } catch (err) {
+            if (err.code === 11000) {
+                const entity_ref: core.Entity_Reference = {uuid: entity_metadata.uuid, kind: entity_metadata.kind};
+                const [metadata, spec] = await this.get_spec(entity_ref);
+                return [metadata, spec, err]
```
adolgarev: think here should throw specialized error type with fields that capture conflicting metadata&spec from db

__tests__/database_mongo.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    });
+    let provider_uuid = uuid4();
+    test("Register Provider", done => {
+        if (providerDb === undefined) {
+            done.fail(new Error("specDb is undefined"));
+            return;
+        }
+        const test_kind = {} as Kind;
+        const provider: Provider = {"uuid": provider_uuid, prefix: "test", version: "0.1", kinds: [test_kind]};
+        providerDb.register_provider(provider).then(res => {
```
adolgarev: I think one can just return promise and jest will resolve it

src/databases/spec_db_interface.ts
```diff
@@ -11,11 +11,11 @@ export interface Spec_DB {
     // implementation needs to CAS the spec_version to the increment
     // of itself, and return the new metadata with the new
     // spec_version and the new CASed in spec.
-    update_spec(entity_metadata: core.Metadata, spec:core.Spec):[boolean, core.Metadata, core.Spec];
+    update_spec(entity_metadata: core.Metadata, spec:core.Spec): Promise<[core.Metadata?, core.Spec?, Error?]>;
```
shlomi-nx: Why is `Metadata` and `Spec` optional? Is this to account for when we have an error without an entity?
I suspect this may be slightly confusing, as a completely empty result is also valid. I'd have metadata and spec be always returned, if they dont exist then a null would indicate that they are missing. 

What do you think?

src/databases/spec_db_mongo.ts
shlomi-nx: I agree

src/databases/spec_db_interface.ts
shlomi-nx: Thinking about this further, I think we need to decide: either we use Exceptions (which I think is the right way), or returned error values. This particular case seems rather confusing, as there are errors that are simply errors (eg line 41 in the implementation), and errors that are returned as a tuple.

Perhaps a better approach would be exactly what you described below: a specific error type that captures the conflict.

src/papiea.ts
```diff
@@ -65,6 +65,8 @@ export interface Procedural_Signature {
 
 // [[file:~/work/papiea-js/Papiea-design.org::#h-Provider-description-189][provider-desc-struct]]
 export interface Provider {
+    // Not sure that we need this, however src/databases/provider_db_interface.ts delete_provider requires this
+    uuid: string;
```
shlomi-nx: See above, I was thinking that the tuple `[prefix, version]` could be the unique identifier of a provider

src/databases/status_db_interface.ts
```diff
@@ -4,26 +4,26 @@ import * as papiea from "../papiea";
...
 
     // Update the status in the status db. As long as the input is
     // correct this always succeeds.
-    update_status(entity_ref:core.Entity_Reference, status:core.Status):boolean;
+    update_status(entity_ref: core.Entity_Reference, status: core.Status): Promise<void>;
 
     // Gets the status of a particular entity from the db. Returns
     // both current metadata and status of the entity.
-    get_status(entity_ref:core.Entity_Reference):core.Status;
+    get_status(entity_ref: core.Entity_Reference): Promise<[core.Metadata?, core.Status?]>;
```
shlomi-nx: Lets discuss Metadata/Status being optional.

src/databases/spec_db_interface.ts
Emulebest: `core.Metadata?` meaning that is could be either null, undefined or metadata. I could make it more verbose like `core.Metadata | null` if that solves the problem. The same goes for spec

shlomi-nx: I think the main question is, can we have only `metadata` and not `spec`? How about `spec` without `metadata`? 

I think that a better signature may be `Promise<[Metadata, Spec] | null>` allowing error cases to be thrown.

Any thoughts?

src/core.ts
```diff
@@ -37,7 +37,7 @@ export interface Metadata {
 
     // Additional fields
     created_at: Date;
-    delete_at: Date;
+    delete_at: Date|null;
```
shlomi-nx: So, could we describe this as `Date?` instead?

src/databases/provider_db_mongo.ts
```diff
@@ -0,0 +1,60 @@
...
+    collection: Collection;
+
+    constructor(db: Db) {
+        this.collection = db.collection("provider");
+    }
+
+    async init(): Promise<void> {
+        try {
+            await this.collection.createIndex({
+                "uuid": 1
```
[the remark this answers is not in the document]
shlomi-nx: What about this point?

src/databases/spec_db_interface.ts
shlomi-nx: I think the question is, can we have only `metadata` and not `spec`? How about `spec` without `metadata`? 

I think that a better signature may be `Promise<[Metadata, Spec] | null>` allowing error cases to be thrown.

Any thoughts?

src/databases/provider_db_mongo.ts
```diff
@@ -0,0 +1,60 @@
...
+        const provider: Provider | null = await this.collection.findOne(filter);
+        if (provider == null) {
+            throw new Error(`Provider with prefix ${provider_prefix} not found`);
+        } else {
+            return provider;
+        }
+    }
+
+    async list_providers(): Promise<Provider[]> {
+        return await this.collection.find({}).toArray();
```
shlomi-nx: Just a comment, I am not sure we require to use `await` here. Since this is already a `Promise`, it is likely that `await` will be used at the call-site. This means that the program wont have to wait until the results are ready (as it is in this implementation), rather it will return a promise and when the returned value is to be used, only then will the program wait.

Its not critical in here, but it is something worth noting. If this explanation made no sense, please let me know so we could discuss this together.

PR #46, opened 2018-12-31
35 Initial implementation of Entity Spec Only CRUD

### LINKED ISSUE #35: CRUD Engine:  Generate CRUD for `spec-only` entities
This will generate the CRUD as described in the [wiki page](https://github.com/nutanix/papiea-js/wiki/Version-0.1-Plan#1-simple-case-for-spec-only-entity).

Essentially, CRUD operations that are trivial database operations, no intentful stuff.

Emulebest:
Resolves #35


src/core.ts
```diff
@@ -11,7 +11,7 @@ export type Data_Description = any;
 
 // Lets define a type for a version. For now it may be string, but could be more
 // elaborate later on
-export type Version = string;
+export type Version = number;
```
shlomi-nx: Not sure a number is general enough for versioning. Why not String?

src/entity/entity_api_interface.ts
```diff
@@ -0,0 +1,16 @@
+import { Kind } from "../papiea";
+import { Metadata, Spec, uuid4, Version } from "../core";
+
+export interface IEntityAPI {
+    get_kind(prefix: string, kind: string): Promise<Kind>
+
+    save_entity(kind: Kind, spec_description: any, status_description?: any): Promise<[Metadata, Spec]>
```
shlomi-nx: spec/status_description should be of types `Spec` and `Status` respectively

src/entity/entity_api_interface.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    get_entity_spec(kind: Kind, entity_uuid: uuid4): Promise<[Metadata, Spec]>
+
+    filter_entity_spec(kind: Kind, fields: any): Promise<[Metadata, Spec][]>
+
+    update_entity_spec(uuid: uuid4, version: Version, kind: Kind, spec_description: any): Promise<[Metadata, Spec]>
```
shlomi-nx: Same here

src/entity/entity_routes.ts
```diff
@@ -0,0 +1,54 @@
...
+        }
+    });
+
+
+    router.get("/:prefix/:kind/:uuid", kind_middleware, asyncHandler(async (req, res) => {
+        const [metadata, spec] = await entity_api.get_entity_spec(req.params.entity_kind, req.params.uuid);
+        res.json({"metadata": metadata, "spec": spec});
+    }));
+
+    // This could be GET /:prefix/:kind with query params as filter
```
shlomi-nx: Agree, perhaps we could expose both options. Post may be used when the filter string is too long for query params.

src/entity/entity_routes.ts
```diff
@@ -0,0 +1,54 @@
...
+    router.post("/:prefix/:kind/filter", kind_middleware, asyncHandler(async (req, res) => {
+        const result = await entity_api.filter_entity_spec(req.params.entity_kind, req.body.filter_fields);
+        res.json({"result": result});
+    }));
+
+    router.put("/:prefix/:kind/:uuid/:version", kind_middleware, asyncHandler(async (req, res) => {
```
shlomi-nx: Interesting, did you mean `spec_version` here? I like this

src/core.ts
Emulebest: The problem is that Metadata interface requires `spec_version: number`, while the Version type requires a String. Moreover we are using a mongodb increment when updating the spec, thus it requires a number to do this (or parsing string as a number). We should settle down with going either string or number everywhere

src/entity/entity_api_interface.ts
Emulebest: Understood

src/entity/entity_api_interface.ts
Emulebest: Understood

src/entity/entity_routes.ts
Emulebest: I'll do both variants

src/entity/entity_routes.ts
Emulebest: Yes, I meant `spec_version` here. It might be required because of our "optimistic locking" mechanism which we provider by incrementing the spec_version on update. Basically it would mean that you specified a concrete version of the entity you are working on. Otherwise we might happen to be in inconsistent state in case of 2 or more people working on the same entity concurrently

shlomi-nx: That is why we use a `cas` operation when updating spec, so that in case of concurrent operations only one will succeed. It may not be a bad idea I am just not yet sure how it helps.

If we give an example for this, it may look something like:
```
err = http.put('/papiea/kind/"+entity.metadata.uuid+"/"+entity.metadata.spec_version, entity);
```

instead of:
```
err = http.put('/papiea/kind/"+entity.metadata.uuid, entity);
```
conveying the same information and the same concurrency guarantees. Unless I am missing something it seems that the second option is much cleaner.

Thoughts?

src/core.ts
adolgarev: these are 2 different versions. (1) version of the spec, this should be a number (integer to be precise) as we increment it and implement in this way optimistic locking, (2) version of the provider, this should be a string, the common format for this is for instance https://docs.npmjs.com/about-semantic-versioning which is a string and cannot be represented as a number.

src/entity/entity_api_impl.ts
```diff
@@ -0,0 +1,61 @@
+import {Status_DB} from "../databases/status_db_interface";
+import {Spec_DB} from "../databases/spec_db_interface";
+import {Provider_DB} from "../databases/provider_db_interface";
+import {Kind} from "../papiea";
+import {Entity, Entity_Reference, Metadata, Spec, Status, uuid4, Version} from "../core";
+import uuid = require("uuid");
+import {IEntityAPI} from "./entity_api_interface";
```
adolgarev: please adhere to the same formatting

src/entity/entity_api_impl.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    async update_entity_spec(uuid: uuid4, version: Version, kind: Kind, spec_description: any): Promise<[Metadata, Spec]> {
+        const metadata: Metadata = {uuid: uuid, kind: kind.name, spec_version: version} as Metadata;
+        return this.spec_db.update_spec(metadata, spec_description);
+    }
+
+    async delete_entity_spec(kind: Kind, entity_uuid: uuid4): Promise<void> {
+        const entity_ref: Entity_Reference = {kind: kind.name, uuid: entity_uuid};
+        await this.spec_db.delete_spec(entity_ref);
+        await this.status_db.delete_status(entity_ref);
```
adolgarev: spec and status are in the same doc, we should set deleted_at in metadata here and do not remove spec&status

src/entity/entity_api_interface.ts
```diff
@@ -0,0 +1,16 @@
+import {Kind} from "../papiea";
+import {Metadata, Spec, uuid4, Version} from "../core";
+
+export interface IEntityAPI {
```
adolgarev: @shlomi-nx @Emulebest we should stick with some naming convention, I'm ok with any of them, but it should be one

src/entity/entity_routes.ts
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+
+    router.get("/:prefix/:kind/:uuid", kind_middleware, asyncHandler(async (req, res) => {
+        const [metadata, spec] = await entity_api.get_entity_spec(req.params.entity_kind, req.params.uuid);
+        res.json({"metadata": metadata, "spec": spec});
+    }));
+
+    // This could be GET /:prefix/:kind with query params as filter
+    router.post("/:prefix/:kind/filter", kind_middleware, asyncHandler(async (req, res) => {
```
adolgarev: https://stackoverflow.com/questions/26104394/rest-api-filter-operator-best-practice

src/entity/entity_api_interface.ts
shlomi-nx: Agree that we should choose one and stick to it. Typescript seems to be firmly against I-prefix for some reason, and it does not use is internally as well. In your experience, what is the commonly used convention amongst typescript developers?

src/entity/entity_routes.ts
adolgarev: we do not send here metadata, the request now looks like:

screenshot_20190110_145124

So we either (1) send metadata in request body OR (2) specify spec_version as url part

src/entity/entity_api_interface.ts
adolgarev: Not sure about typescript but I adhere to following:
* interface needs to have a simple name depicting what it is about like SpecDB - it says that this is a storage for specs
* concrete implementation of the interface must be specific in something that differs it from all other implementations, like SpecDBMongo - this is spec storage in mongo, - the name should be long and ugly, if you see such name in the code then this code smells - one should use interface instead of concrete implementation
* in cases where there is only one possible implementation of the interface I prefer to not introduce an interface at all. Suffixes like Impl are not welcome, for instance, now we have Provider_API and Provider_API_Impl, we could avoid interface here altogether. If in future we needed another implementation I'd introduce interface or abstract class.

adolgarev: Also we should stick with either CamelCase, under_scores, or Mixed_Style for classes/variables.

src/entity/entity_routes.ts
shlomi-nx: metadata must be sent with the request body, yes.

src/entity/entity_api_interface.ts
shlomi-nx: Ohhh, that what you meant.. Sure, lets do Mixed_Style for classes/types and under_scores for variables

src/entity/entity_routes.ts
shlomi-nx: This is a true purist view, but this real-world limitation has been around for a while. Even if we start with `GET` only, we will eventually have to implement a `POST` end point as well, as the limitation will reveal itself sooner or later.

Please see https://stackoverflow.com/a/4203837/1027568 and other answers there for reference.

__tests__/entity_api.test.ts
```diff
@@ -0,0 +1,138 @@
...
+            spec: {
+                x: 10,
+                y: 11
+            }
+        };
+        const spec_query = {
+            q: JSON.stringify(spec)
+        };
+        try {
+            const res = await entityApi.get(`${providerPrefix}/${kind_name}/filter?${stringify(spec_query)}`,);
```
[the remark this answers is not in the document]
shlomi-nx: Yes, this is great for `GET`, we could worry about `POST` later on, but we will have to make it at some point. At that time we should probably have a `filter` url part to denote "unstored `filter` object with a response which is the results", or in other words, find the loophole in the REST way to accommodate with http's query-params length limitation.

__tests__/entity_api.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    test("Update entity spec", async (done) => {
+        expect.assertions(1);
+        try {
+            const res = await entityApi.put(`/${providerPrefix}/${kind_name}/${entity_metadata.uuid}`, {
+                spec: {
+                    x: 20,
+                    y: 21
+                },
+                spec_version: 1
```
adolgarev: this should be
```
metadata: {
  spec_version: 1
}
```

Issue #59, opened 2019-01-31
non-static "procedure" is done at SDK level instead of at Kind level


shlomi-nx:
I am going over the code, and I found the following https://github.com/nutanix/papiea-js/blob/9fd48531a0903ea916562bee5d48fece075aad0f/src/provider_sdk/typescript_sdk_interface.ts#L11-L23 which puts procedure at the provider level, but "specializes" it with `specified_kind_name?`. 

If we want provider-level procedures, then it should not be specialized like that (I think we decided to hold off on such procedures for now). If we want a procedure on an instance of a kind, than we should only use `kind.procedure` as specified in https://github.com/nutanix/papiea-js/blob/9fd48531a0903ea916562bee5d48fece075aad0f/src/provider_sdk/typescript_sdk_interface.ts#L39-L43

I am referring to this portion of the location example: https://github.com/nutanix/papiea-js/blob/4d02be7920ac350098b546eeaa4c57bec5b4d85c/examples/typescript/location/provider/provider.ts#L21-L29 I think this should be something like:

```ts
    const loc = sdk.new_kind(cfg.provider.kind_description);
    //...
    //Registers a procedures on papiea instance
    loc.procedure(proceduralSignature.name, {}, proceduralSignature.execution_strategy, proceduralSignature.argument, proceduralSignature.result,  async (ctx, entity, input) => {
        entity.spec.x += input;
        const res = await axios.put(ctx.url_for(entity), {
            spec: entity.spec,
            metadata: entity.metadata
        });
        return res.data;
    });

```
WDYT?

PR #65, opened 2019-02-01
58_Custom metadata additional info field 

### LINKED ISSUE #58: Custom fields in metadata at entity creation time

Emulebest:
Resolves #58 #59

### DIFF
--- a/__tests__/provider_sdk.test.ts
+++ b/__tests__/provider_sdk.test.ts
@@ -2,9 +2,9 @@ import "jest"
 import { load } from "js-yaml";
 import { readFileSync } from "fs";
 import { resolve } from "path";
-import { ProviderSdk } from "../src/provider_sdk/typescript_sdk";
+import { Kind_Procedure_Builder, ProviderSdk } from "../src/provider_sdk/typescript_sdk";
 import { plural } from "pluralize"
-import { Kind, Procedural_Execution_Strategy, Procedural_Signature } from "../src/papiea";
+import { Procedural_Execution_Strategy, Procedural_Signature } from "../src/papiea";
 import { loadYaml } from "./test_data_factory";
 import axios from "axios"
 
@@ -57,8 +57,8 @@ describe("Provider Sdk tests", () => {
     });
     test("Provider can create a new kind", (done) => {
         const sdk = ProviderSdk.create_sdk(papiea_config.host, papiea_config.port, server_config.host, server_config.port);
-        const kind = sdk.new_kind(location_yaml);
-        expect(kind.name).toBe("Location");
+        const location_manager = sdk.new_kind(location_yaml);
+        expect(location_manager.kind.name).toBe("Location");
         done();
     });
     test("Provider with no x-papiea-entity should fail", (done) => {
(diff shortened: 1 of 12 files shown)


__tests__/provider_sdk.test.ts
```diff
@@ -112,18 +112,18 @@ describe("Provider Sdk tests", () => {
         sdk.new_kind(geo_location_yaml);
         done();
     });
-    let location_kind: Kind;
+    let location_kind_manager: Kind_Procedure_Builder;
```
adolgarev: it can be not only procedure builder but just kind builder, kind may have another stuff to be defined

src/databases/spec_db_mongo.ts
```diff
@@ -23,12 +23,14 @@ export class Spec_DB_Mongo implements Spec_DB {
     }
 
     async update_spec(entity_metadata: core.Metadata, spec: core.Spec): Promise<[core.Metadata, core.Spec]> {
+        const filter: any = {};
+        for (let key in entity_metadata) {
+            if (key !== "created_at" && key !== "deleted_at") {
+                filter["metadata." + key] = entity_metadata[key];
```
adolgarev: don't think additional fields should be in a filter, they should be in update

src/entity/entity_api_impl.ts
```diff
@@ -3,7 +3,7 @@ import { Status_DB } from "../databases/status_db_interface";
 import { Spec_DB } from "../databases/spec_db_interface";
 import { Provider_DB } from "../databases/provider_db_interface";
 import { Kind, Procedural_Signature } from "../papiea";
-import { Data_Description, Entity_Reference, Metadata, Spec, uuid4 } from "../core";
+import { AdditionalData, Data_Description, Entity_Reference, Metadata, Spec, uuid4 } from "../core";
```
adolgarev: AdditionalData type is not needed

src/core.ts
```diff
@@ -14,6 +14,10 @@ export type Data_Description = any;
 export type Version = string;
 // core-types ends here
 
+export type AdditionalData = {
```
adolgarev: AdditionalData type is not needed

src/entity/entity_api_impl.ts
```diff
@@ -40,10 +40,17 @@ export class EntityAPI implements EntityApiInterface {
         return found_kind;
     }
 
-    async save_entity(kind: Kind, spec_description: Spec): Promise<[Metadata, Spec]> {
-        const metadata: Metadata = { uuid: uuid(), spec_version: 0, created_at: new Date(), kind: kind.name };
+    async save_entity(kind: Kind, spec_description: Spec, request_metadata: Metadata = {} as Metadata): Promise<[Metadata, Spec]> {
+        if (!request_metadata.uuid) {
```
adolgarev: check uuid to be a valid uuid

__tests__/provider_sdk.test.ts
shlomi-nx: Absolutely agree. This is an "SDK" representation of a kind, meaning in here we define what this kind will actually be able to do (such as procedures, validations, intentful operations, dependencies etc). 

I was imagining this being simply exposed with the name `Kind` on the SDK, but if this causes a clash, we could potentially refer to it as a `Kind_Builder` or something of that sort. Remember that this type only exists on the SDK.

PR #66, opened 2019-02-06
Integration fixes

adolgarev:


__tests__/nunet/host_type.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    const dataDescription: Data_Description = loadYaml("./nunet/host_type_data.yml");
+    const kindName = "HostType";
+    beforeAll(async () => {
+        try {
+            const { data } = await entityApi.get(`/${providerPrefix}/${kindName}`);
+            for (var i = 0; i < data.length; i++) {
+                await entityApi.delete(`/${providerPrefix}/${kindName}/${data[i].metadata.uuid}`);
+            }
+            await providerApi.delete(`${providerPrefix}/${providerVersion}`);
+        } catch (e) {
```
Emulebest: I guess putting something like `throw new Error("Unreachable")` will be useful

src/databases/spec_db_mongo.ts
```diff
@@ -89,15 +98,15 @@ export class Spec_DB_Mongo implements Spec_DB {
...
+                $set: {
+                    "metadata.deleted_at": new Date()
+                }
+            });
         if (result.result.n === undefined || result.result.ok !== 1) {
             throw new Error("Failed to remove spec");
         }
         if (result.result.n !== 1 && result.result.n !== 0) {
-            throw new Error(`Amount of entities deleted must be 0 or 1, found: ${ result.result.n }`);
+            throw new Error(`Amount of entities deleted must be 0 or 1, found: ${result.result.n}`);
```
Emulebest: Idk why, but seems your linter omits spaces in format strings. Let's agree on this piece of code style

__tests__/entity_api.test.ts
```diff
@@ -98,15 +98,21 @@ describe("Entity API tests", () => {
...
-            const res = await entityApi.post(`${ providerPrefix }/${ kind_name }/filter`, {
-                filter_fields: {
-                    spec: {
-                        x: 10,
-                        y: 11
-                    }
+            let res = await entityApi.post(`${ providerPrefix }/${ kind_name }/filter`, {
+                spec: {
+                    x: 10,
+                    y: 11
```
shlomi-nx: What about partial matches?

```typescript
let res = await entityApi.post(`${ providerPrefix }/${ kind_name }/filter`, {
                spec: {
                    x: 10
                }
            });
```

Should return all entities that have in their spec `x: 10`, no matter what `y` is. We should also be able to make partial or complete queries on status and metadata as well.

We should add a test for that.

__tests__/nunet/geolocation_data.yml
```diff
@@ -0,0 +1,35 @@
...
+      required:
+        - latitude
+        - longitude
+      properties:
+        latitude:
+          type: number
+          example: 38.06539529956805
+        longitude:
+          type: number
+          example: -79.35421246073054
```
shlomi-nx: Nunet specific stuff should not be in papiea's repo, not even as tests. This part is not open sourced and should remain out of the public domain

__tests__/nunet/geolocation_list_v3.json
```diff
@@ -0,0 +1,94 @@
...
+                "provider": "AWS",
+                "provider_region": "eu-west-1",
+                "location": {
+                    "longitude": -6.223781624163166,
+                    "latitude": 53.4050427791716
+                }
+            }
+        }
+    ]
+}
```
shlomi-nx: Same goes here, this needs to be removed..

__tests__/nunet/host_type_data.yml
```diff
@@ -0,0 +1,22 @@
...
+    name:
+      type: string
+    provider:
+      type: string
+    instance_type:
+      type: string
+    cpu_model:
+      type: string
+    memory_capacity_mib:
+      type: number
```
shlomi-nx: Same here, please remove.

__tests__/nunet/host_type_list_v3.json
```diff
@@ -0,0 +1,101 @@
...
+            "status": {
+                "name": "t2.xlarge",
+                "provider": "AWS",
+                "instance_type": "t2.xlarge",
+                "cpu_model": "Intel Xeon E5 2686 v4",
+                "memory_capacity_mib": 204800
+            }
+        }
+    ]
+}
```
shlomi-nx: And here. Please remove.

__tests__/nunet/spec_only_entity.test.ts
```diff
@@ -0,0 +1,172 @@
...
+                }
+            });
+            expect(res.data.length).toBe(0);
+        } catch (e) {
+            done.fail(e);
+            return;
+        }
+        done();
+    });
+});
```
shlomi-nx: This tests should not "belong" to nunet. We should convert these to some made up provider.

__tests__/entity_api.test.ts
adolgarev: We do partial matching, that is it will indeed return all entities that have in their spec `x: 10`, no matter what `y` is.
How do you think one can express that he wants exact match via API?

shlomi-nx: I am not too worried about exact matches. If it will become relevant we may expose a "find" endpoint or something which has to be an exact match.

I was more concerned with not seeing a test case for such scenario.

PR #69, opened 2019-02-13
Integrate nunet

litvaOo:


litvaOo:
Needs to update `.circleci/Dockerfile` to use `papiea:latest` from jfrog if everything is okay, because there is no such image at the moment.

PR #73, opened 2019-02-20
Stat papiea faster from prebuild image

### LINKED ISSUE #71: Papiea is taking a while to start in docker container
I see Papiea's clojure stuff being compiled long after the entire application (nunet) is already up. 

We should fix the container creation of papiea to do all the compilations at build time. At runtime it should start immediately.

adolgarev:
Resolves #71

### DIFF
--- a/.circleci/Dockerfile
+++ b/.circleci/Dockerfile
@@ -1,26 +1,11 @@
-FROM nutanix-docker.jfrog.io/papiea:latest
+FROM nutanix-docker.jfrog.io/papiea:base
 
 COPY . /code
-
 WORKDIR /code
+RUN cd /code \
+    # Use cached node_modules
+    && mv /node_modules /code/node_modules \
+    && npm install \
+    && npm run build-clj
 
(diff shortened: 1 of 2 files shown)


adolgarev:
@shlomi-nx
* nutanix-docker.jfrog.io/papiea:base is a base image which is built by docker-compose.yml from Dockerfile in the root folder, this image contains all dependencies (ubuntu packages, clojure compiler, closure and node modules) but doesn't contain source code, changes rarely, I pushed it manually
* CI uses base image, adds there a source code, compiles, runs tests and pushes resulting image with tag nutanix-docker.jfrog.io/papiea:latest , without base image it will take a long time to build such an image from scratch each time CI runs.
* you are getting an error above because .circleci/docker-compose.yml must be moved to the root folder first, see circle ci config how it is done, CI build passes successfully

shlomi-nx:
Thanks that was very helpful. Merging to master, will test tomorrow.

PR #76, opened 2019-02-25
Status of spec-only entities should be equal to the spec

### LINKED ISSUE #75: Get entity on spec-only entities should duplicate spec as status

adolgarev:
Resolves #75

### DIFF
--- a/src/entity/entity_routes.ts
+++ b/src/entity/entity_routes.ts
@@ -45,7 +45,8 @@ export function createEntityRoutes(entity_api: EntityAPI): Router {
 
     router.get("/:prefix/:kind/:uuid", kind_middleware, asyncHandler(async (req, res) => {
         const [metadata, spec] = await entity_api.get_entity_spec(req.params.entity_kind, req.params.uuid);
-        res.json({ "metadata": metadata, "spec": spec });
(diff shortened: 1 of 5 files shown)


src/entity/entity_routes.ts
```diff
@@ -45,7 +45,8 @@ export function createEntityRoutes(entity_api: EntityAPI): Router {
 
     router.get("/:prefix/:kind/:uuid", kind_middleware, asyncHandler(async (req, res) => {
         const [metadata, spec] = await entity_api.get_entity_spec(req.params.entity_kind, req.params.uuid);
-        res.json({ "metadata": metadata, "spec": spec });
+        const [_, status] = await entity_api.get_entity_status(req.params.entity_kind, req.params.uuid);
+        res.json({ "metadata": metadata, "spec": spec, "status": status });
```
Emulebest: I see that we are including status in the response as we discussed on the meeting. The question is, aren't we supposed to get status of each entity while querying all of the entities too?

shlomi-nx: Yes, if we are querying for multiple entities, each should contain its full view, composed of meta+spec+status

adolgarev: @Emulebest @shlomi-nx see https://github.com/nutanix/papiea-js/pull/78

PR #79, opened 2019-03-04
Golang Provider SDK

### LINKED ISSUE #54: Generate Golang Client SDK based on Swagger Model

Emulebest:
resolves #54 
This is WIP

### DIFF
--- a/src/provider_sdk/golang/common/sdk_impl.go
+++ b/src/provider_sdk/golang/common/sdk_impl.go
@@ -1,7 +1,6 @@
 package common
 
 import (
-	"../papiea"
 	"fmt"
 	"github.com/jinzhu/inflection"
 	"github.com/pkg/errors"
@@ -11,16 +10,16 @@ import (
 )
 
 type ProviderSdk struct {
-	Version       *papiea.ProviderVersion
+	Version       *ProviderVersion
 	Prefix        *string
-	Kinds         []papiea.Kind
-	Provider      *papiea.Provider
+	Kinds         []Kind
+	Provider      *Provider
 	PapieaUrl     string
 	PapieaPort    int
 	ServerManager ServerManager
 }
 
-func makeSdk(hostParams map[string]string) (*ProviderSdk, error) {
+func MakeSdk(hostParams map[string]string) (*ProviderSdk, error) {
 	papieaHost, present := hostParams["papiea_host"]
 	if present == false {
 		return nil, errors.New("Papiea host is not specified")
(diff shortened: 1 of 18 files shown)


src/provider_sdk/golang/papiea/papiea.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type Spec interface {
+	SpecOf() string
+}
+
+type Entity struct {
+	Metadata *Metadata `json:"metadata,omitempty"`
+	Spec     *Spec     `json:"spec,omitempty"`
+	Status   *Status   `json:"status,omitempty"`
```
shlomi-nx: For these entities, we would kinda like to avoid pointers for now. It is harder to print and debug, and makes reasoning slightly more difficult. 

Unless there is a good reason, lets try to avoid pointers for now.

Emulebest: We might need pointers on spec & status because of `omitempty` to surpass openapi check of empty field

src/provider_sdk/golang/common/sdk_impl.go
```diff
@@ -0,0 +1,182 @@
...
+		} else {
+			return nil, errors.New("Couldn't interpret papiea port as number")
+		}
+		return nil, errors.New("Papiea port is not specified")
+	}
+	sdk.PapieaUrl = papieaHost
+	router, routerGroup := makeRouter()
+	serverManager := ServerManager{
+		"127.0.0.1",
+		9000,
```
adolgarev: why host/port is hardcoded here?

src/provider_sdk/golang/papiea/papiea.go
```diff
@@ -0,0 +1,83 @@
...
+	HaltIntentful ExecutionStrategy = 0
+)
+
+type ProviderPower int
+
+const (
+	On        ProviderPower = 0
+	Off       ProviderPower = 1
+	Suspended ProviderPower = 2
+)
```
adolgarev: can we somehow generate a bunch of this by openapi spec?

src/provider_sdk/golang/common/sdk_impl.go
Emulebest: Though it is hardcoded here, we check if function params contain optional public host & public port (lines 52, 54) and override that

src/provider_sdk/golang/papiea/papiea.go
Emulebest: If we have a finite list of states I'd better have them as enums, however I am not sure that we do have a finite list?

Emulebest:
It is irrelevant ATM

PR #93, opened 2019-04-04
Implement RBAC + ABAC using casbin

### LINKED ISSUE #86: Implement configurable OAUTH2
Papiea should be able to configure authentication using oauth2. 

Auth may be disabled as well

### LINKED ISSUE #87: Implement RBAC + ABAC using casbin
To do #18 we could use casbin.

### LINKED ISSUE #88: Provider registration should specify RBAC model and policy
Using casbin each provider could specify its own model and url for policy. 

We should also have some default ones available for quickly starting.

adolgarev:
Resolves #86 
Resolves #87 
Resolves #88


shlomi-nx:
This looks quite good. Some important things to keep in mind:

1. There should be a default authz model, maybe what we currently have
2. However, a provider should be able to register his own model/policy if the default one is not sufficient.
3. When integrating this with the oauth module, some request headers should be set by inspecting the returned oauth token. Since these headers should also be customizable by the provider writers, I turned that into a little DSL in the oauth config file. Please review https://github.com/nutanix/papiea-js/pull/94 and in particular https://github.com/nutanix/papiea-js/blob/349b4baa8516f4d23a6e1b51acba8a39aa6b7116/__tests__/auth.yaml. The "^" prefix denotes a lookup in the environment, and a string that does not start with "^" is simply a constant.
4. So the headers that should be looked at by the authz part should also be configurable, and should match the headers being set by the auth.yaml's `user_data` above.

PR #103, opened 2019-04-15
Added SDK support for kind & provider level procedures

Emulebest:


src/provider_sdk/typescript_sdk.ts
```diff
@@ -110,9 +112,40 @@ export class ProviderSdk implements ProviderImpl {
         this._prefix = prefix;
     }
 
+    provider_procedure(name: string, rbac: any,
+                       strategy: Procedural_Execution_Strategy,
+                       input_desc: any,
+                       output_desc: any,
+                       handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<any>): void {
+        const callback_url = this.server_manager.callback_url(name);
```
adolgarev: Urls should be different for provider, kind and entity level procedures, in the kind and entity case url should include kind name

PR #108, opened 2019-04-18
Added patch method for status & some tests

Emulebest:


__tests__/database_mongo.test.ts
```diff
@@ -150,6 +150,29 @@ describe("MongoDb tests", () => {
...
         done();
     });
+    test("Partially update Status", async (done) => {
+        const statusDb: Status_DB = await connection.get_status_db();
+        const entity_ref: core.Entity_Reference = { uuid: entityA_uuid, kind: "test" };
+        const status: core.Status = { b: "A3" };
+        await statusDb.partial_update_status(entity_ref, status);
+        done();
+    });
+    test("Get partially updated Status", async (done) => {
```
adolgarev: `Partially update Status` & `Get partially updated Status` should be one test, tests could be run independently

src/databases/status_db_interface.ts
```diff
@@ -25,6 +25,8 @@ export interface Status_DB {
     list_status(fields_map: any): Promise<([core.Metadata, core.Status])[]>;
 
     delete_status(entity_ref: core.Entity_Reference): Promise<void>
+
+    partial_update_status(entity_ref: core.Entity_Reference, status: core.Status): Promise<void>
```
adolgarev: * think should stick with better naming like replace_status & update_status where the latter actually does partial update. The same in EntityAPI
* add comments yo the API methods

PR #111, opened 2019-04-22
Add docker configuration for development

### LINKED ISSUE #110: Provide development server docker-compose config
Please provide a docker-compose-dev.yml that will have a nodemon hot-reload dev server inside

litvaOo:
resolves #110


docker-compose.yml
```diff
@@ -10,6 +10,8 @@ services:
                         - mongo
                 ports:
                         - "3333:3000"
+                environment:
+                        - HOT_RELOAD=${HOT_RELOAD}
```
adolgarev: if variable is not set will it give you a warning? is there a way to not get a warning in this case?

PR #113, opened 2019-04-24
Fix generation of api doc for provider procedures

litvaOo:


adolgarev:
@litvaOo add there PATCH for spec/status as well

Emulebest:
@adolgarev spec doesn't have a PATCH method. All the status manipulation methods should be features in provider facing api docs

Issue #134, opened 2019-05-22
Invalid swagger yaml on procedures with empty arg/ret


shlomi-nx:
If no argument/return shapes are defined for procedures (or defined as nothing `{}`), we get invalid swagger yaml:

```yaml
    post:
      description: Calls a procedure ping
      operationId: callxi_clustersping
      tags:
        - xi_clusters/0.0.1/imaged_host/procedure
      parameters:
        - name: uuid
          in: path
          description: UUID of the entity
          required: true
          schema:
            type: string
            format: uuid
      requestBody:
        description: ping input
        required: true
        content:
          application/json:
            schema:
              properties:
                input:
                  $ref: '#/components/schemas/undefined'    # <=
      responses:
        '200':
          description: ping response
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/undefined'  # <=
```

Example code which may generates this is: 
```js
    //                                                here----v---v
    imaged_hosts.entity_procedure('ping', {}, proc_strategy, {}, {}, ping_handler)
```

shlomi-nx:
@Emulebest Please take a look, see whats missing. If nothing you can close.

Emulebest:
resolved via #177

PR #147, opened 2019-05-24
Fixed undefined in procedures that didn't specify validation

Emulebest:
resolves #147

### DIFF
--- a/papiea-engine/__tests__/api_docs.test.ts
+++ b/papiea-engine/__tests__/api_docs.test.ts
@@ -129,47 +129,4 @@ describe("API Docs Tests", () => {
     test("API Docs should be accessible by the url", done => {
         api.get("/api-docs/api-docs.json").then(() => done()).catch(done.fail);
     });
-
-    test("Provider with procedures that contain no validation scheme should be {}", async done => {
-        const procedure_id = "computeSumNoValidation";
-        const proceduralSignatureForProvider: Procedural_Signature = {
-            name: "computeSumWithNoValidation",
-            argument: loadYaml("./procedure_sum_input.yml"),
-            result: loadYaml("./procedure_sum_output.yml"),
-            execution_strategy: Procedural_Execution_Strategy.Halt_Intentful,
-            procedure_callback: "127.0.0.1:9010"
-        };
-        const provider: Provider = new ProviderBuilder("provider_no_validation_scheme")
-                            .withVersion("0.1.0")
-                            .withKinds()
-                            .withCallback(`http://127.0.0.1:9010`)
-                            .withProviderProcedures({ [procedure_id]: proceduralSignatureForProvider })
-                            .build();
-        console.dir(provider);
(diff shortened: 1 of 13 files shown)


shlomi-nx:
This is related to #134 but not exactly it.

I did not mean to create unvalidated procedures, rather, procedures which validates they get no input or return no output. Like a `void` type.

I am not entirely sure I want to keep `unvalidated` procedures around, I much rather them get validated to `any object` (https://swagger.io/docs/specification/data-models/data-types/#object look for `free-form` object)

Emulebest:
> This is related to #134 but not exactly it.
> 
> I did not mean to create unvalidated procedures, rather, procedures which validates they get no input or return no output. Like a `void` type. The following three permutations are needed:
> 
> 1. `(void)->a`
> 2. `(a)->void`
> 3. `(void)->void`
> 4. `(a)->b` - what we currently support
> 
> I am not entirely sure I want to keep `unvalidated` procedures around, if someone needs this, I much rather the procedures get validated to `any object` (https://swagger.io/docs/specification/data-models/data-types/#object look for `free-form` object)

Regarding this one, I defaulted all the procedures that didn't specify a validation scheme to `AnyType` which is basically is an `any` type. Seems like I chose naming for this stuff incorrectly

shlomi-nx:
I have seen what you have did, and that was my comment exactly. Any is assignable from anything, so this is a type which is always valid. That is not what I want, I want a type like `Nothing` that fails on everything.

papiea-engine/__tests__/api_docs.test.ts
```diff
@@ -123,4 +129,47 @@ describe("API Docs Tests", () => {
...
+            await providerApi.post('/', provider);
+            const apiDoc = await apiDocsGenerator.getApiDocs();
+            console.dir(apiDoc.paths);
+            expect(apiDoc.paths[`/services/${ provider.prefix }/${ provider.version }/procedure/${ procedure_id }`]
+                .post
+                .requestBody
+                .content["application/json"]
+                .schema
+                .properties
+                .input['$ref']).toEqual(`#/components/schemas/AnyValue`);
```
shlomi-nx: This should be `NoValue` which represents the absence of any value.

papiea-engine/src/api_docs/api_docs_generator.ts
```diff
@@ -291,8 +291,18 @@ export default class ApiDocsGenerator {
         };
     }
 
+    processEmptyValidation(proc_def: any, sig: Procedural_Signature) {
+        if (!sig.argument) {
+            proc_def.requestBody.content["application/json"].schema.properties.input.$ref = `#/components/schemas/AnyValue`
+        }
+        if (!sig.result) {
+            proc_def.responses["200"].content["application/json"].schema.properties.input.$ref = `#/components/schemas/AnyValue`
```
shlomi-nx: Same here

Issue #131, opened 2019-05-22
Add tests for generated swagger.json


shlomi-nx:
Lets add tests to swagger.json:

1. Create a simple provider with all three kinds of procedures (entity, kind, provider) with input/output models.
2. Test that swagger.json emits proper fields for all CRUD info on that kind 
   - Also make sure to test the structure of the input/output of each
3. Test that swagger.json emits proper fields for all Procedures

shlomi-nx:
Also make sure to add a test case that checks that no extra arguments are required, as in this ticket #133

shlomi-nx:
@Emulebest Please take a look, see whats missing. If nothing you can close.

Emulebest:
resolved via #177

PR #177, opened 2019-06-12
Api docs tests & status fields excluded from api docs

### LINKED ISSUE #161: Swagger - status-only fields are presented as spec fields
If we have an kind such as:
```yaml
test_kind:
  type: object
  x-papiea-entity: spec-only
  properties:
    status_field:
      x-papiea: status-only
      type: string
```

In the swagger ui for `POST` (creation) I am seeing:
```json
{
  "spec": {
    "status_field": "string"
   }
}
```

Instead of an empty spec (in this case)

Emulebest:
resolves #161, #134, #131


papiea-engine/__tests__/api_docs.test.ts
```diff
@@ -6,37 +6,31 @@ import { loadYaml, ProviderBuilder } from "./test_data_factory";
 import { Provider_DB } from "../src/databases/provider_db_interface";
 import { Provider, Version, Procedural_Signature, Procedural_Execution_Strategy } from "papiea-core";
 import ApiDocsGenerator from "../src/api_docs/api_docs_generator";
+
 declare var process: {
     env: {
+        MONGO_DB: string,
+        MONGO_HOST: string,
+        MONGO_PORT: string
```
adolgarev: please remove this

PR #192, opened 2019-06-27
Fix logout issue

shlomi-nx:
resolves #286

Lets try to revoke refresh_token as well.


shlomi-nx:
This was solved in nunet-ui. We will need to make sure login/logout works well for other oauth providers when this becomes relevant. 

See fixes here https://github.com/nutanix/nunet-ui/pull/360

PR #199, opened 2019-07-09
Manage s2s keys by uuid

litvaOo:
resolves https://github.com/nutanix/nunet-ui/issues/345


papiea-engine/src/databases/s2skey_db_mongo.ts
```diff
@@ -32,11 +32,19 @@ export class S2S_Key_DB_Mongo implements S2S_Key_DB {
         return;
     }
 
-    async get_key(key: Key): Promise<S2S_Key> {
-        const result: S2S_Key | null = await this.collection.findOne({
-            "key": key,
-            "deleted_at": null
-        });
+    async get_key(key: Key, name?: string): Promise<S2S_Key> {
```
shlomi-nx: This function definition is very hacky - why is `key` mandatory if one passes `name`? it is completely unused.

Please split this into `get_key(key:Key)` and `get_key_by_name(name:string)`

papiea-engine/src/databases/s2skey_db_mongo.ts
```diff
@@ -32,11 +32,19 @@ export class S2S_Key_DB_Mongo implements S2S_Key_DB {
...
+        let result: S2S_Key | null;
+        if (name) {
+            result = await this.collection.findOne({
+                "name": name,
```
shlomi-nx: what ensures our name is unique? `findOne` simply gets the first match, but if we have a few with the same name (think `test`), this will fail.

papiea-engine/src/databases/s2skey_db_mongo.ts
litvaOo: Done

papiea-engine/src/databases/s2skey_db_mongo.ts
litvaOo: Added index on 'name' field

shlomi-nx:
@adolgarev I agree with your points. @litvaOo Please change the implementation of `get_key_by_name` accordingly. It should probably be `get_keys_by_name` and it should return all the keys with the same name that the user may access.

litvaOo:
@shlomi-nx 
I do not think we need a separate `get_keys_by_name` method to return list of keys with corresponding name, because we already have `list_keys` method, which accepts `field_map` for search

papiea-engine/src/provider/provider_routes.ts
```diff
@@ -53,7 +53,7 @@ export default function createProviderAPIRouter(providerApi: Provider_API) {
     }));
 
     providerApiRouter.get('/:prefix/:version/s2skey', asyncHandler(async (req, res) => {
-        const s2skeys = await providerApi.list_keys(req.user, {});
+        const s2skeys = await providerApi.list_keys(req.user, {"deleted_at": req.query.deleted});
```
adolgarev: what is this for?

litvaOo: This is needed to return only keys, that are not deleted, or if user needs - deleted at certain date.

adolgarev:
> do not see changes to package.json, thus please do not commit your local changes to package-lock.json
@litvaOo ^^^

litvaOo:
@adolgarev 
I think it is better to always commit package-lock, because we can use it to increase build time with `npm ci`
But if needed, I can revert this particular changes

adolgarev:
> @adolgarev
> I think it is better to always commit package-lock, because we can use it to decrease build time with `npm ci`
> But if needed, I can revert this particular changes

there is a package-log.json already, you just update it without a reason

litvaOo:
@adolgarev 
Reverted changes to package-lock files, think we can merge it

PR #212, opened 2019-07-12
Call procedure auth tests

adolgarev:
* If one uses a policy
`p, alice, owner, ${kind_name}, *, allow`
this means that owner of the entity may do all actions on top of it, including all procedure calls
* To be more explicit
```
p, alice, owner, ${kind_name}, read, allow
p, alice, owner, ${kind_name}, callmovex, allow
````
This allows alice to call moveX entity procedure for entities of which she is an owner, note that read permission is also required as entity will be read and given to the procedure
* auth for kind and provider level procedures does not work with current casbin model for provider-user because there is no owner or tenant on kind or provider level
* provider-admin can call all kind and provider procedures of his provider

### DIFF
--- a/papiea-engine/src/auth/authz.ts
+++ b/papiea-engine/src/auth/authz.ts
@@ -35,7 +35,7 @@ export const ReadAction = new Action('read'),
     InactivateS2SKeyAction = new Action('inactivate_key');
 
 export function CallProcedureByNameAction(procedureName: string) {
-    return new Action('call' + procedureName.toLowerCase());
+    return new Action('call_' + procedureName.toLowerCase());
 }
 
 function mapAsync<T, U>(array: T[], callbackfn: (value: T, index: number, array: T[]) => Promise<U>): Promise<U[]> {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: papiea-engine/src/auth/casbin.ts [hunk 1]
```diff
@@ -24,7 +24,12 @@ export class CasbinAuthorizer extends Authorizer {
     }
 
     async checkPermission(user: UserAuthInfo, object: any, action: Action): Promise<void> {
-        if (!this.enforcer.enforce(user, object, action.getAction())) {
+        try {
+            if (!this.enforcer.enforce(user, object, action.getAction())) {
+                throw new PermissionDeniedError();
+            }
+        } catch (e) {
+            console.error(e);
```
shlomi-nx: Logs without any text are really hard to track down. Until we use a proper logger that shows where the log message came from, lets have some inline text on logs messages.

papiea-engine/src/auth/authz.ts
```diff
@@ -35,7 +35,7 @@ export const ReadAction = new Action('read'),
     InactivateS2SKeyAction = new Action('inactivate_key');
 
 export function CallProcedureByNameAction(procedureName: string) {
-    return new Action('call' + procedureName);
+    return new Action('call' + procedureName.toLowerCase());
```
shlomi-nx: For readability of the policies, lets turn this into `call_` procedure name

PR #231, opened 2019-07-17
Generate random secret on each run

### LINKED ISSUE #219: Generate new secret on each papiea run

Emulebest:
resolves #219

### DIFF
--- a/papiea-engine/src/utils/utils.ts
+++ b/papiea-engine/src/utils/utils.ts
@@ -1,4 +1,6 @@
 import { ValidationError } from "../validator";
+import "js-sha256"
+import { sha256 } from "js-sha256";
 
 export class Maybe<T> {
     private constructor(private value: T | null) {}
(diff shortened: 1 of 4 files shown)


papiea-engine/src/utils/utils.ts
```diff
@@ -71,4 +73,19 @@ export function isEmpty(obj: any) {
...
+    for(let i = 0; i < len; i++) {
+        random_ascii = Math.floor((Math.random() * 25) + 97);
+        random_string += String.fromCharCode(random_ascii)
+    }
+    return random_string
+}
+
+export function generateSecret(): string {
+    let rnd = generateRandomString(50);
+    return sha256(rnd);
```
adolgarev: Consider https://security.stackexchange.com/a/109626 , but take into account that author confuses urandom and random, see https://en.wikipedia.org/wiki//dev/random . Anyways given a good random source you do not need hashing. Read https://paragonie.com/blog/2016/05/how-generate-secure-random-numbers-in-various-programming-languages#nodejs-csprng about what is a good random source for nodejs (not Math.random!). If you use a bad random source like Math.random you need to add time before hashing

Emulebest: So you'd suggest adding salt in the form of time?

adolgarev: at least

shlomi-nx:
@Emulebest If you are done with this and this is still required, please merge

Emulebest:
Since there is no notion of papiea token now this is not needed anymore

PR #239, opened 2019-07-25
Procedures check security

### LINKED ISSUE #227: Security: ProviderSDK needs to make an auth check to papiea when executing procedures
When the providerSdk is receiving a procedure call, it gets the `headers` from papiea (needs #225). It then needs to make a call back to papiea, to ensure that the received token is in-fact allowed to execute this function. Otherwise, some third party may be able to send commands to run procedures on providers without properly authenticating.

Emulebest:
resolves #227

### DIFF
--- a/papiea-engine/src/auth/authz.ts
+++ b/papiea-engine/src/auth/authz.ts
@@ -1,27 +1,12 @@
 import { UserAuthInfo, UnauthorizedError } from "./authn";
 import { Provider_API } from "../provider/provider_api_interface";
-import { Provider } from "papiea-core";
+import { Provider, Action } from "papiea-core";
 
 export class PermissionDeniedError extends Error {
     constructor() {
         super("Permission Denied");
         Object.setPrototypeOf(this, PermissionDeniedError.prototype);
     }
-}
-
-export enum Actions {
-    ReadAction = "read",
-    UpdateAction = "write",
-    CreateAction = "create",
-    DeleteAction = "delete",
-    RegisterProvider = "register_provider",
-    UnregisterProvider = "unregister_provider",
-    ReadProvider = "read_provider",
-    UpdateAuth = "update_auth",
-    CreateS2SKey = "create_key",
-    ReadS2SKey = "read_key",
-    InactivateS2SKey = "inactive_key",
-    UpdateStatus = "update_status",
 }
 
 function mapAsync<T, U>(array: T[], callbackfn: (value: T, index: number, array: T[]) => Promise<U>): Promise<U[]> {
@@ -37,9 +22,9 @@ export abstract class Authorizer {
     constructor() {
     }
 
-    abstract checkPermission(user: UserAuthInfo, object: any, action: Actions): Promise<void>;
+    abstract checkPermission(user: UserAuthInfo, object: any, action: Action): Promise<void>;
 
-    async filter(user: UserAuthInfo, objectList: any[], action: Actions, transformfn?: (object: any) => any): Promise<any[]> {
+    async filter(user: UserAuthInfo, objectList: any[], action: Action, transformfn?: (object: any) => any): Promise<any[]> {
         return filterAsync(objectList, async (object) => {
             try {
                 if (transformfn) {
@@ -56,7 +41,7 @@ export abstract class Authorizer {
 }
 
 export class NoAuthAuthorizer extends Authorizer {
-    async checkPermission(user: UserAuthInfo, object: any, action: Actions): Promise<void> {
+    async checkPermission(user: UserAuthInfo, object: any, action: Action): Promise<void> {
     }
 }
 
@@ -119,7 +104,7 @@ export class PerProviderAuthorizer extends Authorizer {
         return authorizer;
     }
 
-    async checkPermission(user: UserAuthInfo, object: any, action: Actions): Promise<void> {
+    async checkPermission(user: UserAuthInfo, object: any, action: Action): Promise<void> {
         const authorizer: Authorizer | null = await this.getAuthorizerByObject(user, object);
         if (authorizer === null) {
             return;
@@ -143,8 +128,8 @@ export class PerProviderAuthorizer extends Authorizer {
 }
 
 export class AdminAuthorizer extends Authorizer {
-    async checkPermission(user: UserAuthInfo, object: any, action: Actions): Promise<void> {
-        if (action === Actions.ReadProvider) {
+    async checkPermission(user: UserAuthInfo, object: any, action: Action): Promise<void> {
+        if (action === Action.ReadProvider) {
             return;
         }
         if (!user) {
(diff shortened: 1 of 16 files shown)


papiea-engine/src/auth/authz.ts
```diff
@@ -9,31 +9,21 @@ export class PermissionDeniedError extends Error {
...
-    private action: string;
-
-    constructor(action: string) {
-        this.action = action;
-    }
-
-    getAction(): string {
-        return this.action;
-    }
+export enum Actions {
```
adolgarev: what for changed class to Enum? Anyways do not make it plural, it serves as a type and  declaration `action : Actions` is not right as it says that argument action represents many actions, it should be `action : Action`
BTW rename then ReadAction to Read, etc

papiea-engine/src/entity/entity_routes.ts
```diff
@@ -31,6 +32,25 @@ export function createEntityAPIRouter(entity_api: Entity_API): Router {
         return {results: pageEntities, entity_count: totalEntities};
     };
 
+    router.post("/:prefix/:version/check_permission", asyncHandler(async (req, res) => {
+        const allowed: boolean = await entity_api.check_permission(req.user, req.params.prefix, req.params.version, req.body.action, req.body);
```
adolgarev: api call MUST check permission, the check here is redundant

papiea-sdk/src/provider_sdk/typescript_sdk_context_impl.ts
```diff
@@ -52,14 +89,28 @@ export class ProceduralCtx implements ProceduralCtx_Interface {
...
-            if (parts[0] === 'Bearer') 
+            const parts = this.headers.authorization.split(' ');
+            if (parts[0] === 'Bearer')
                 return parts[1]
         }
         throw new Error("No invoking user")
     }
+}
+
+export enum Actions {
```
adolgarev: why not to move this to core then and do not duplicate the code?

papiea-engine/src/entity/entity_api_interface.ts
```diff
@@ -22,4 +24,8 @@ export interface Entity_API {
     call_kind_procedure(user: UserAuthInfo, prefix: string, kind_name: string, version: Version, procedure_name: string, input: any): Promise<any>
 
     call_provider_procedure(user: UserAuthInfo, prefix: string, version: Version, procedure_name: string, input: any): Promise<any>
+
+    check_permission(user: UserAuthInfo, prefix: string, version: Version, action: Actions, entityRef: Entity_Reference): Promise<boolean>
+
+    check_permissions(user: UserAuthInfo, prefix: string, version: Version, action: Actions, requestParams: any): Promise<boolean>
```
adolgarev: what is requestParams?

papiea-engine/src/auth/authz.ts
Emulebest: If this a class, we can't transfer this as string param. I am not sure we had this as a class in the first place

papiea-engine/src/entity/entity_api_interface.ts
Emulebest: either `metadata` or `entity_ref`

shlomi-nx: `Metadata` is an `Entity_Reference` by implementation. If anything, this type could have been `Metadata[]`. However, I think this actually needs to be a list of `[Action, Metadata]` tuple, this way, the singular case is just a degenerate form where the list is of size 1.

papiea-engine/src/entity/entity_routes.ts
shlomi-nx: @adolgarev Did not quite get your message, this check is exactly what this api is doing - checking that these operations are allowed.

papiea-engine/src/entity/entity_routes.ts
```diff
@@ -31,6 +32,25 @@ export function createEntityAPIRouter(entity_api: Entity_API): Router {
...
+    }));
+
+    router.post("/:prefix/:version/check_permissions", asyncHandler(async (req, res) => {
+        const allowed: boolean = await entity_api.check_permissions(req.user, req.params.prefix, req.params.version, req.body.action, req.body);
+        if (allowed) {
+            res.json({"success": "Ok"})
+        } else {
+            throw new PermissionDeniedError()
+        }
+        console.log()
```
shlomi-nx: Why is this here?

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -22,50 +22,50 @@ class SecurityApiImpl implements SecurityApi {
     readonly provider: ProviderSdk;
     readonly s2s_key: string;
     constructor (provider:ProviderSdk, s2s_key:Key) {
-        this.provider = provider
+        this.provider = provider;
```
shlomi-nx: Just as a side note, I see you keep adding `;` to the end of lines. I personally try to avoid that, I do not like that style. We have gone a long way from c ;) Don't feel compelled to put these everywhere, actually, you are encouraged to remove them.

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -410,6 +410,7 @@ export class Kind_Builder {
         const version = this.get_version();
         this.server_manager.register_handler(`/${this.kind.name}/${name}`, async (req, res) => {
             try {
+                console.log(req.headers);
```
shlomi-nx: Please add contextual messages to log prints

papiea-sdk/src/provider_sdk/typescript_sdk_utils.ts
```diff
@@ -1,3 +1,6 @@
+import { Metadata } from "papiea-core";
+import { Entity_Reference } from "papiea-core/build/core";
```
shlomi-nx: wrong `from` clause. This should just be added to the line above.

papiea-sdk/src/provider_sdk/typescript_sdk_interface.ts
```diff
@@ -63,6 +65,7 @@ export interface IntentfulCtx_Interface {
     get_user_security_api(user_s2skey:Key): SecurityApi
     get_headers(): IncomingHttpHeaders
     get_invoking_token(): Key
+    check_permission(provider_prefix: string, provider_version: Version, entity: any, action: Actions): Promise<boolean>
```
shlomi-nx: please remove `prefix` and `version`. They are to be picked up from the context. 

If we will want to add the ability to check permissions for other providers we will add that explicitly.

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -454,6 +454,7 @@ export class Kind_Builder {
...
```
shlomi-nx: @Emulebest Please stop pushing logs with no strings for context.

Emulebest: Sorry, will double check for debug logs

PR #243, opened 2019-07-31
Move prefix and owner to root

### LINKED ISSUE #224: s2s key db collection should have provider_prefix, owner, is_admin, is_provider_admin in the root
~It should only have it in the `extension` section.~

Lets have s2skey's `user_info` as similar as possible to the `user_info` retrieved by analyzing idp tokens. Lets move `provider_prefix` OUT of `user_info` into the root, as well as `owner` and `is_provider_admin` and `is_admin`.

litvaOo:
resolves #224


papiea-engine/src/auth/authz.ts
```diff
@@ -138,14 +138,14 @@ export class AdminAuthorizer extends Authorizer {
...
+                || object.is_admin) {
                 throw new PermissionDeniedError();
             }
             if (user.is_provider_admin) {
                 return;
             }
-            if (object.userInfo.is_provider_admin
-                || object.userInfo.owner !== user.owner) {
+            if (object.is_provider_admin
+                || object.owner !== user.owner) {
```
adolgarev: this is not enough as we do a lot of checks involving `user_info.is_admin`, `user_info.is_provider_admin` and `user_info.provider_prefix`. Thus you need to have is_admin, is_provider_admin and provider_prefix info in user_info object or pass it alongside user_info object
Please revise provider_admin_auth.test.ts - it still passes is_admin and is_provider_admin in user_info and not as req.body.is_admin, req.body.is_provider_admin and because of this tests pass

PR #245, opened 2019-07-31
Added 'Secret' as a class & changed naming

### LINKED ISSUE #242: Lets get Key type back
I dont know why type `Key` was removed in this commit https://github.com/nutanix/papiea-js/commit/0dd59f84df19e0c2e05a1fc0bb5c2a62b46a7926

It is much more typesafe to designate that this is a `Key` even though its an alias to a string.

Please undo this change, and lets discuss these kinds of changes next time.

Emulebest:
resolves #242


papiea-core/src/core.ts
```diff
@@ -194,6 +194,14 @@ export interface UserInfo {
     [key: string]: any;
 }
 
+export interface Secret<T> {
+    _secret: T
```
shlomi-nx: In any case, a private member should not be defined in an interface

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -51,26 +52,26 @@ class SecurityApiImpl implements SecurityApi {
...
+            const {data: r } = await this.provider.provider_api_axios.put(`${url}/s2skey`, {uuid: uuid, active:false}, {headers: {'Authorization': `Bearer ${this.secret.getSecret()}`}});
             return r
         } catch (e) {
             console.log("error getting s2skeys", e);
             throw new Error("Cannot get s2skeys" + e.message)
         }
     }
 }
-export class ProviderSdk implements ProviderImpl {
+export class ProviderSdk<T> implements ProviderImpl<T> {
```
shlomi-nx: Same here.

papiea-sdk/src/provider_sdk/typescript_sdk_crypto.ts
```diff
@@ -0,0 +1,17 @@
+import { Secret } from "papiea-core";
+
+export class SecretImpl<T> implements Secret<T> {
```
shlomi-nx: This definition needs to move to core. Should not be repeated.

papiea-sdk/src/provider_sdk/typescript_sdk.ts
Emulebest: agree

papiea-sdk/src/provider_sdk/typescript_sdk_crypto.ts
Emulebest: I thought that we have a general interface in core, leaving impl to the invoking side. I should rethink this thin wrapper

papiea-core/src/core.ts
```diff
@@ -194,19 +200,11 @@ export interface UserInfo {
...
-
-    setSecret(secret: T): void
-}
-
 export interface S2S_Key {
     name?: string
     owner: string
     provider_prefix: string
-    key: string
+    secret: any
```
shlomi-nx: Then this should be the `Secret` type

papiea-engine/__tests__/auth_tests/entity_api_auth.test.ts
```diff
@@ -331,9 +331,9 @@ describe("Entity API auth tests", () => {
             { headers: { 'Authorization': 'Bearer ' + token } }
         );
         expect(s2skeys.length).toEqual(1);
-        expect(s2skeys[0].key).toEqual(s2skey.key.slice(0, 2) + "*****" + s2skey.key.slice(-2));
+        expect(s2skeys[0].secret).toEqual(s2skey.secret.slice(0, 2) + "*****" + s2skey.secret.slice(-2));
         const { data } = await providerApi.get(`/${ provider.prefix }/${ provider.version }/auth/user_info`,
-            { headers: { 'Authorization': 'Bearer ' + s2skey.key } }
+            { headers: { 'Authorization': 'Bearer ' + s2skey.secret } }
```
shlomi-nx: `s2skey.secret.getSecret()`

papiea-core/src/core.ts
```diff
@@ -185,6 +185,12 @@ export interface Provider {
     allowExtraProps: boolean;
 }
 
+export interface Secret {
+    getSecret(): any
+
+    setSecret(secret: any): void
```
shlomi-nx: I would do this differently:
```typescript
export interface Secret {
   getSecret(): any
}

export function create_string_secret(string secret) {
   return {getSecret : ()=>secret}
}
```

papiea-engine/__tests__/auth_tests/provider_api_auth.test.ts
```diff
@@ -64,19 +64,19 @@ describe("Provider API auth tests", () => {
     test("Admin should create s2s key for provider-admin with key provided", async () => {
         jest.setTimeout(5000);
         expect.hasAssertions();
-        const key = uuid();
+        const secret = uuid();
```
shlomi-nx:
```typescript
const secret = create_string_secret(uuid())
```

papiea-engine/__tests__/auth_tests/provider_api_auth.test.ts
```diff
@@ -64,19 +64,19 @@ describe("Provider API auth tests", () => {
...
                     provider_prefix: provider.prefix,
                     is_provider_admin: true
                 }
             }
         );
-        expect(s2skey.key).toEqual(key);
+        expect(s2skey.secret).toEqual(secret);
         const { data: userInfo } = await providerApi.get(`/${ provider.prefix }/${ provider.version }/auth/user_info`,
-            { headers: { 'Authorization': `Bearer ${ s2skey.key }` } }
+            { headers: { 'Authorization': `Bearer ${ s2skey.secret }` } }
```
shlomi-nx:
```typescript
{ headers: { 'Authorization': `Bearer ${ s2skey.secret.getSecret() }` } }
```

papiea-engine/src/auth/crypto.ts
```diff
@@ -7,18 +7,18 @@ export function createHash(obj: any): string {
...
 }
 
-export class SecretImpl<T> implements Secret<T> {
-    _secret: T
+export class SecretImpl implements Secret {
+    private secret: any
 
-    constructor(secret: T) {
-        this._secret = secret
+    constructor(secret: any) {
```
shlomi-nx: With my proposal, we do not need such class, just the constructor shown above.

papiea-engine/src/databases/s2skey_db_interface.ts
```diff
@@ -6,7 +6,7 @@ export interface S2S_Key_DB {
 
     get_key(uuid: string): Promise<S2S_Key>;
 
-    get_key_by_secret<T>(secret: T): Promise<S2S_Key>;
+    get_key_by_secret(secret: any): Promise<S2S_Key>;
```
shlomi-nx: use `Secret` type here

papiea-engine/src/main.ts
```diff
@@ -48,7 +49,7 @@ async function setUpApplication(): Promise<express.Express> {
     const statusDb = await mongoConnection.get_status_db();
     const s2skeyDb = await mongoConnection.get_s2skey_db();
     const providerApi = new Provider_API_Impl(providerDb, statusDb, s2skeyDb, new AdminAuthorizer());
-    app.use(createAuthnRouter(adminKey, s2skeyDb, providerDb));
+    app.use(createAuthnRouter(new SecretImpl(adminKey), s2skeyDb, providerDb));
```
shlomi-nx: Here you could just pass the secret constructor function

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -38,10 +38,10 @@ class SecurityApiImpl<T> implements SecurityApi<T> {
         }
     }
 
-    public async list_keys(): Promise<T[]>{
+    public async list_keys(): Promise<S2S_Key[]>{
         try {
             const url = `${this.provider.get_prefix()}/${this.provider.get_version()}`;
-            const { data: keys } = await this.provider.provider_api_axios.get(`${url}/s2skey`, {headers: {'Authorization': `Bearer ${this.secret.getSecret()}`}});
+            const {data: keys } = await this.provider.provider_api_axios.get(`${url}/s2skey`, {headers: {'Authorization': `Bearer ${this.secret.getSecret()}`}});
```
shlomi-nx: Good, thats what it all should look like

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -286,12 +286,12 @@ export class ProviderSdk<T> implements ProviderImpl<T> {
         throw new Error(`Malformed provider description. Missing: ${ missing_field }`)
     }
 
-    static create_provider<T>(papiea_url: string, secret: T, public_host?: string, public_port?: number, allowExtraProps: boolean = false): ProviderSdk<T> {
+    static create_provider(papiea_url: string, secret: any, public_host?: string, public_port?: number, allowExtraProps: boolean = false): ProviderSdk {
```
shlomi-nx: use `Secret`

papiea-sdk/src/provider_sdk/typescript_sdk_crypto.ts
```diff
@@ -1,17 +1,17 @@
 import { Secret } from "papiea-core";
 
-export class SecretImpl<T> implements Secret<T> {
-    _secret: T
+export class SecretImpl implements Secret {
```
shlomi-nx: This is not required, and in any case must not be duplicated. `Secret` definition along with its set of constructors should all be in `papiea-core`

papiea-core/src/core.ts
Emulebest: If this is to be a `Secret` type, then it won't be serializable in JSON

papiea-sdk/src/provider_sdk/typescript_sdk_crypto.ts
Emulebest: I am not sure that implementations like `SecretImpl` or helper functions like `create_string_secret` should live in Core

papiea-engine/src/auth/crypto.ts
Emulebest: It's more about functional vs OOP approach IMO. We have mostly everything in OOP approach, why don't we stick to that?

papiea-core/src/core.ts
shlomi-nx: Thats ok to have custom code which will serialize/deserialize it. Also, thats one of the reasons I originally went with just a type alias. 

However, since this is not merely an internal type, secret should indeed be `Secret`

papiea-sdk/src/provider_sdk/typescript_sdk_crypto.ts
shlomi-nx: Why not?

shlomi-nx: Also, if you look at my suggestions in https://github.com/nutanix/papiea-js/pull/245#discussion_r310289021, this will not even be required

papiea-engine/src/auth/crypto.ts
shlomi-nx: Some times heavy class syntax is more convenient, other times a simple map is. If you look at the client code, you will see a lot of that being used. I don't feel that we need to pick one and stick with it, rather we should use the most appropriate one. In this situation, the secret should denote an immutable structure for holding secrets. I don't see the benefits of having this in a class with subclasses.

adolgarev: @Emulebest 
do not name concrete implementations like SomethingImpl. What differs this implementation from others? Name it accordingly like StringSecret implements Secret

@Emulebest @shlomi-nx 
we also need to think on JSON API we have as we talk to clients not in terms of classes or closures but in terms of JSON. Thus we need to define a JSON scheme client will expect on his end. I tend to avoid custom JSON serialization/deserialization, thus what I see would be a good solution here is following:
```
export interface Secret {
   getSecret(): any
}

class StringSecret implements Secret {
    private readonly key: string;
    
    constructor(key: string) {
        this.key = key;
    }
    
    getSecret(): string {
        return this.key;
    }
}

const secret = new StringSecret("abc");
console.log("secret.getSecret()", secret.getSecret());
console.log("JSON.stringify(secret)", JSON.stringify(secret));
```
```
secret.getSecret() abc
JSON.stringify(secret) {"key":"abc"}
```
which means the code inside papiea may use Secret interface: `secret.getSecret()` - and work ok with any implementation of this interface. On the client side we define that we may have different secrets BUT for a string secret we pass a dict with a 'key' attribute that holds this string secret.

Any thoughts or wisdoms?

Emulebest: Add a `secret.toResponse()` to interface definition because it is difficult to overload default JS serialization

papiea-engine/__tests__/auth_tests/entity_api_auth.test.ts
Emulebest: Agree with @adolgarev

papiea-engine/src/auth/crypto.ts
adolgarev: @Emulebest why do you need `secret.toResponse()`? `JSON.stringify(secret)` gives what you want

papiea-core/src/core.ts
adolgarev: > If we present the type as `secret: Secret` then it needs to be an interface containing only fields without functions.

@Emulebest sorry, didn't get this

> I'd suggest we stick with internal representation of Secret when using JSON interfaces

@Emulebest sorry, didn't get this as well

papiea-engine/src/auth/crypto.ts
shlomi-nx: @adolgarev sounds good. 
@Emulebest Lets proceed with this proposal.

shlomi-nx:
Lets talk about this f2f on Friday. I don't know why we are complicating this so much. It was fine with just being a type alias. Anyway, this is low priority right now.

papiea-core/src/core.ts
```diff
@@ -192,11 +194,37 @@ export interface Provider {
 }
 
 export interface Secret {
+    key: any
```
shlomi-nx: @Emulebest What the purpose of defining an internal field in the interface? Doesn't `getSecret` abstracts over it?

Emulebest: the field `key` is required when serializing to JSON, `getSecret()` is not serializable though

Emulebest:
@shlomi-nx I think we should rethink this completely and make it KISS. Probably using an alias, as you did initially, is the way to go now. We are overengineering this for no practical reason right now

Emulebest:
Moved to a simpler alternative #297

PR #246, opened 2019-08-01
Use winston for logging

### LINKED ISSUE #211: Add logging framework to papiea
Lets use some proper logging framework that accepts logging levels and add logs to the system.

litvaOo:
resolves #211


shlomi-nx:
@adolgarev Regarding your last point, why do we want to have loggers pass around in such a way? Why should we not use some default logger, which should probably be set up once at system start, and if we really needs to pass some special logger, do so only then?

papiea-engine/src/databases/mongo.ts
```diff
@@ -38,42 +39,42 @@ export class MongoConnection {
         return this.client.close(true);
     }
 
-    async get_spec_db(): Promise<Spec_DB_Mongo> {
+    async get_spec_db(logger: Logger): Promise<Spec_DB_Mongo> {
```
adolgarev: pass logger to the MongoConnection constructor?

papiea-engine/src/logger.ts
```diff
@@ -0,0 +1,115 @@
...
+                new winston.transports.Console()
+            ],
+        });
+    }
+
+    private isValidLevel(logLevel: string) {
+        return this.logLevels.includes(logLevel)
+    }
+
+    public setLoggingLevel(logLevel: string) {
```
adolgarev: do we need to change log level on the fly?

papiea-engine/src/logger.ts
```diff
@@ -0,0 +1,115 @@
...
+
+function safeJSONParse(chunk: string) {
+    try {
+        return JSON.parse(chunk);
+    } catch (e) {
+        return undefined;
+    }
+}
+
+export class Logger {
```
adolgarev: Logger should be an interface or abstract class, WinstonLogger should implement Logger interface

papiea-engine/src/logger.ts
litvaOo: I wasn't sure about this, so decided to add this method. If we know for sure, that we won't need it, I can remove this part

papiea-engine/src/databases/mongo.ts
litvaOo: We might need use different loggers for different DBs, so I thought it would be better to pass them separately and not with MongoConnection constructor

adolgarev:
> @adolgarev Regarding your last point, why do we want to have loggers pass around in such a way? Why should we not use some default logger, which should probably be set up once at system start, and if we really needs to pass some special logger, do so only then?

there are 2 common approaches:
* factory method: `Logger logger = Logger.getLogger(Mailer.class);`, - here every component imports factory method and asks to create a logger based on component name or class, this is useful when you do not have proper dependency injection in place
* inject logger as a dependency from outside, for instance, in constructor, see how this is done in java world https://medium.com/simars/inject-loggers-in-spring-java-or-kotlin-87162d02d068 , what we are doing is exactly second option - inject logger at constructor

shlomi-nx:
@adolgarev Also, perhaps, if we are already threading a `logger` between all of our functions, why wont we generalize and pass a `context` which would be a map containing various contextual stuff, such as a logger?

adolgarev:
> @adolgarev Also, perhaps, if we are already threading a `logger` between all of our functions, why wont we generalize and pass a `context` which would be a map containing various contextual stuff, such as a logger?

@shlomi-nx not a good idea, usually when you have too many stuff being passed around (like you pass to many things in a class constructor) then this means that class does too many unrelated things and must be broken into smaller ones with better cohesion. (1) Moving a map around just hides this problem and (2) one never knows what is in there, at least context should be strictly typed

shlomi-nx:
@adolgarev I agree with strictly typing a context. Ok, lets do that if we need to. 

@litvaOo Please fix the 2 points and we can merge this.

PR #248, opened 2019-08-02
User info cleanup

### LINKED ISSUE #224: s2s key db collection should have provider_prefix, owner, is_admin, is_provider_admin in the root

adolgarev:
Resolves #224 

* provider_prefix for userInfo from idp is not required
* for s2skeys provider_prefix is mandatory
* do not allow to set it in the request body when creating s2s key, papiea sets it by itself to the proper value


papiea-engine/src/provider/provider_api_impl.ts
```diff
@@ -130,8 +130,9 @@ export class Provider_API_Impl implements Provider_API {
             key: "",
             created_at: new Date(),
             deleted_at: undefined,
-            userInfo: userInfo ? userInfo : user
+            userInfo: {"provider_prefix": provider_prefix}
```
shlomi-nx: Why do have both this definition and the one on line 129?

adolgarev: Do you mean why do we s2skey.owner and s2skey.user_info.owner? First one says who is the owner of the key, second one says what will be the owner when this key is used. When ordinary user creates a s2skey s2skey.owner always equals s2skey.user_info.owner (see AdminAuthorizer which checks that object.owner === user.owner during s2skey creation), but provider-admin and papiea admin may create keys which will allow to talk on behalf of other user

shlomi-nx: @adolgarev No, why we have `s2skey.userinfo` and `s2skey.provider_prefix.userinfo`.

Also, it should be named `user_info`, not `userInfo`

adolgarev: The same applies to provider_prefix, s2skey.provider_prefix says to which provider this s2skey belongs, s2skey.user_info.provider_prefix says to which provider user on behalf of whom you are talking when using this s2skey can access. It is always the same currently, but I left the possibility to have them not equal. Should I remove one?

adolgarev: It is updated to latest master and now is user_info indeed

shlomi-nx: @adolgarev Ok, so maybe we should improve the name, it is very confusing at the moment.

adolgarev: addressed in last commit

PR #260, opened 2019-08-08
Added session token to reduce the auth request data size

### LINKED ISSUE #259: Token received by user needs to be a session token that points to db
We can't persist the original tokens as cookies as they are way too large.

Instead, when a user successfully logs in, we need to establish an app-session for that login, which needs to have a similar expiry as the access token. The id of the session needs to be a key into a db table containing the original token (or the parsed UserInfo since it will not change for that specific session).

When a user makes requests into papiea, he can either use s2s-keys or a session-id. Papiea will check the session-id against the db, see if it needs to renew the token using the refresh token, and return the UserInfo of the user for further processing.

The db will need to be periodically cleaned up to remove expired sessions entries.

When a user logs out, we need to clear that session and remove the db entry for that session. 

An attempt to access papiea with an invalid session needs to be redirected to login.

Emulebest:
Resolves #259


papiea-engine/__tests__/auth_tests/entity_api_auth.test.ts
```diff
@@ -205,7 +207,7 @@ describe("Entity API auth tests", () => {
         await entityApi.delete(`/${provider.prefix}/${provider.version}/${kind_name}/${entity_metadata.uuid}`);
     });
 
-    test("Get user info", async () => {
+    test.only("Get user info", async () => {
```
adolgarev: only?

papiea-engine/package-lock.json
```diff
@@ -5736,613 +5736,6 @@
         "pluralize": "^7.0.0",
         "swagger-model-validator": "^3.0.6",
         "wait-port": "^0.2.2"
-      },
```
adolgarev: package.json has not changed, please do not push your package-lock.json

papiea-engine/src/auth/oauth2.ts
```diff
@@ -90,11 +76,8 @@ export function createOAuth2Router(logger: Logger, redirect_uri: string, provide
     }));
 
     router.use('/provider/:prefix/:version/auth/logout', asyncHandler(async (req, res) => {
-        const provider: Provider = await providerDb.get_provider(req.params.prefix, req.params.version);
-        const oauth2 = getOAuth2(provider);
-        const token = oauth2.accessToken.create({ "access_token": req.user.authorization.split(' ')[1] });
         try {
-            await token.revoke('access_token');
```
adolgarev: still need to revoke a token from idp

papiea-engine/src/auth/session_key.ts
```diff
@@ -0,0 +1,65 @@
...
+    async createKey(userInfo: UserAuthInfo, token: any, key: string): Promise<SessionKey> {
+        const parsedToken = parseJwt(token.token.access_token)
+        const exp = parsedToken.content.exp
+        const sessionKey: SessionKey = {
+            key: key,
+            expireAt: new Date(exp * 1000),
+            user_info: userInfo,
+            idpToken: token
+        }
+        if (SessionKeyAPI.isExpired(sessionKey)) {
```
shlomi-nx: We should attempt to use the `refresh_token` to extend the `access_token`'s duration

papiea-engine/src/auth/session_key.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export class SessionKeyAPI {
+    private readonly sessionKeyDb: SessionKeyDb
+
+    constructor(sessionKeyDb: SessionKeyDb) {
+        this.sessionKeyDb = sessionKeyDb
+    }
+
+    async createKey(userInfo: UserAuthInfo, token: any, key: string): Promise<SessionKey> {
+        const parsedToken = parseJwt(token.token.access_token)
```
shlomi-nx: Is `access_token` a common oauth2 thing? Or should we add a portion in our auth.yml dsl to capture that?

papiea-engine/src/auth/session_key.ts
```diff
@@ -0,0 +1,65 @@
...
+    async getKey(key: string): Promise<SessionKey> {
+        const sessionKey = await this.sessionKeyDb.get_key(key)
+        if (!SessionKeyAPI.isExpired(sessionKey)) {
+            return sessionKey
+        } else {
+            throw new Error("Access token expired")
+        }
+    }
+
+    async inActivateKey(key: string) {
```
shlomi-nx: "Inactivate" is a single word verb, it should be `inactivateKey`

papiea-engine/src/auth/session_key.ts
shlomi-nx: https://github.com/nutanix/papiea-js/issues/262

papiea-engine/src/auth/session_key.ts
shlomi-nx: https://github.com/nutanix/papiea-js/issues/263

Emulebest: It is a standard for OpenID

papiea-engine/src/auth/session_key.ts
Emulebest: my bad, will do

PR #269, opened 2019-08-12
Auth enhancements: Refresh tokens, logout fixed, login redirects

### LINKED ISSUE #268: Fix login/logout
1. Logout simply does not work at the moment. We need to point to papiea's logout which will clear cookies, remove session from db and send a 200 response with a manual redirect to the pre-configured logout url (in auth.yml) which the UI will manually redirect to (avoiding CORS)
2. Need to add automatic token refresh
3. When cannot refresh the token and token is invalid, user must be redirected back to login

Emulebest:
resolves #268


papiea-engine/src/auth/authn.ts
```diff
@@ -88,7 +89,13 @@ export function createAuthnRouter(logger: Logger, userAuthInfoExtractor: UserAut
 
         const user_info = await userAuthInfoExtractor.getUserAuthInfo(token, provider_prefix, provider_version);
         if (user_info === null) {
-            throw new UnauthorizedError();
+            if (provider_prefix && provider_version) {
```
adolgarev: 1. authn should not depend on provider db and oauth2
2. why do you do the redirect? one needs to return 401 Unauthorized because these requests are made via XHR and not put as query string in the browser. XHR won't and shouldn't handle the login form. JS sends XHR, receives 401 and goes to login page - `https://${location.host}/provider/xi_clusters/0.0.1/auth/login?redirect_uri=https://${location.host}/`

papiea-engine/src/auth/session_key.ts
```diff
@@ -1,59 +1,88 @@
...
+        }
         return sessionKey
     }
 
-    async getKey(key: string): Promise<SessionKey> {
+    async getKey(key: string, oauth2: any): Promise<SessionKey> {
         const sessionKey = await this.sessionKeyDb.get_key(key)
-        if (!SessionKeyAPI.isExpired(sessionKey)) {
-            return sessionKey
+        if (SessionKeyAPI.isExpired(sessionKey.idpToken)) {
```
adolgarev: `sessionKey.isExpired()` ?

Emulebest: Since entities in Papiea Core are basically DTO and SessionKeyAPI as an analogy to other entites' APIs handles validation (which `isExpired` kind of is) and business logic, I'd suggest we keep it as is to keep everything uniform

papiea-engine/src/auth/session_key.ts
```diff
@@ -1,59 +1,88 @@
 import { SessionKeyDb } from "../databases/session_key_db_interface"
 import { UserAuthInfo, UserAuthInfoExtractor } from "./authn"
 import { SessionKey } from "papiea-core"
-import { parseJwt } from "./user_data_evaluator"
+import { Provider_DB } from "../databases/provider_db_interface"
+import { getOAuth2 } from "./oauth2"
 
 export class SessionKeyAPI {
+    private static EXPIRATION_WINDOW_IN_SECONDS = 300
```
shlomi-nx: Why do we need this constant? Should we not pick it up from the idp provided token itself?

shlomi-nx: https://github.com/nutanix/papiea-js/issues/271

Emulebest: @shlomi-nx 
```
The expired helper is useful for knowing when a token has definitively expired. However, there is a common race condition when tokens are near expiring. If an OAuth 2.0 token is issued with a expires_in property (as opposed to an expires_at property), there can be discrepancies between the time the OAuth 2.0 server issues the access token and when it is received. These come down to factors such as network and processing latency. This can be worked around by preemptively refreshing the access token:
```
Taken from https://www.npmjs.com/package/simple-oauth2

Emulebest: Basically, we take the access token expiration and subtract this constant to not get into the race condition

shlomi-nx: I see what you meant. Lets keep it.

PR #280, opened 2019-08-15
Added Intentful Behaviour layer

### LINKED ISSUE #279: Spec update overrides status-only fields

Emulebest:
resolves #279


papiea-engine/src/intentful_core/intentful_context.ts
```diff
@@ -0,0 +1,28 @@
+import { Spec_DB } from "../databases/spec_db_interface"
+import { Status_DB } from "../databases/status_db_interface"
+import { IntentfulStrategy } from "./intentful_strategy_interface"
+import { BasicIntentfulStrategy } from "./basic_intentful_strategy"
+import { IntentfulBehaviour } from "papiea-core"
+
+export type BehaviourStrategyMap = Map<IntentfulBehaviour, IntentfulStrategy>
+
+export class IntentfulContext {
```
adolgarev: Why do you call this a 'context'? Meaning of context is
> the circumstances that form the setting for an event, statement, or idea, and in terms of which it can be fully understood.

This is a factory https://www.tutorialspoint.com/design_pattern/factory_pattern

Emulebest: I call it context because of future use of this class can have related functionality (creating tasks https://nutanix.github.io/papiea-js/Papiea-design.html#h-Intentful-Tasks-117, etc.)

papiea-engine/src/intentful_core/basic_intentful_strategy.ts
```diff
@@ -0,0 +1,32 @@
...
+    private readonly specDb: Spec_DB
+    private readonly statusDb: Status_DB
+
+    constructor(specDb: Spec_DB, statusDb: Status_DB) {
+        this.specDb = specDb
+        this.statusDb = statusDb
+    }
+
+    // Update spec and status with delta received in spec
+    async update(metadata: Metadata, spec: Spec): Promise<Spec> {
```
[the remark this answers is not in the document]
Emulebest: Ok, I meant 'delta' in the meaning of spec change, need to reformulate it

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -155,24 +155,21 @@ export class ProviderSdk implements ProviderImpl {
...
-                    entity_procedures: {},
-                    differ: undefined,
-                };
-                const kind_builder = new Kind_Builder(spec_only_kind, this, this.allowExtraProps);
-                this._kind.push(spec_only_kind);
-                return kind_builder;
-            } else {
-                //TODO: process non spec-only
-                throw new Error("Unimplemented")
+            const kind: Kind = {
```
shlomi-nx: Where did `spec-only`? There is a distinction here:
- spec-only - should NOT allow to update status, since it essentially has none
- basic entity - every change in a spec field is automatically reflected in the corresponding status field. Deviation from it will have to be manually resolved (someone changed status without changing spec)
- intentful entity - every change is spec field is only reflected in status based on differ logic (**LATER**)

shlomi-nx:
@Emulebest Since we don't have a lot of time at all, meaning, **this must be merged in today and all UI operations have to work with it**, lets not change existing things. Just add the `basic entity` on top of what we currently have.

papiea-engine/src/intentful_core/basic_intentful_strategy.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    // Update spec and status with spec changes received
+    async update(metadata: Metadata, spec: Spec): Promise<Spec> {
+        const [updatedMetadata, updatedSpec] = await this.specDb.update_spec(metadata, spec);
+        const [_, existingStatus] = await this.statusDb.get_status(metadata)
+        if (existingStatus) {
+            await this.statusDb.update_status(metadata, spec)
+        } else {
+            await this.statusDb.replace_status(metadata, spec)
```
adolgarev: I believe for spec-only we should always call replace_status

PR #287, opened 2019-08-22
Changed logger to be pretty printed & fixed a bug and added test

### LINKED ISSUE #294: Pretty print logs in console

Emulebest:
resolves #294 
New logs in console look like the following:
```
{  
   "0":{  
      "two":"errors"
   },
   "1":{  
      "three":"errors"
   },
   "2":{  
      "Time":"2019-08-21T15:00:42.118Z"
   },
   "level":"error",
   "message":{  
      "one":"error"
   }
}
```

### DIFF
--- a/papiea-engine/src/logger.ts
+++ b/papiea-engine/src/logger.ts
@@ -1,7 +1,8 @@
-import * as winston from 'winston';
-import { resolve } from "path";
-import { NextFunction, Request } from "express";
-import Logger from './logger_interface';
+import * as winston from 'winston'
+import { resolve } from "path"
+import { NextFunction, Request } from "express"
+import Logger from './logger_interface'
+import { safeJSONParse } from "./utils/utils"
 
 export function getLoggingMiddleware(logger: Logger) {
     return async (req: Request, res: any, next: NextFunction): Promise<void> => {
@@ -33,14 +34,6 @@ export function getLoggingMiddleware(logger: Logger) {
     }
 }
 
-function safeJSONParse(chunk: string) {
-    try {
-        return JSON.parse(chunk);
-    } catch (e) {
-        return undefined;
-    }
-}
-
 export class WinstonLogger implements Logger {
     private logger: winston.Logger;
     private readonly logLevels = ["emerg", "alert", "crit", "error", "warning", "notice", "info", "debug"];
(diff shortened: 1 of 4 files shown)


papiea-engine/src/logger.ts
```diff
@@ -49,16 +42,22 @@ export class WinstonLogger implements Logger {
         if (!this.isValidLevel(logLevel)) {
```
nkntnx: Should we lose the log just because the log level is incorrect ? I suggest we log that we found an invalid log level, and then use error as the default logging here. That way the flow won't break in case of invalid log level.

papiea-engine/src/logger.ts
```diff
@@ -49,16 +42,22 @@ export class WinstonLogger implements Logger {
...
             transports: [
-                new winston.transports.File({ filename: logFile ?
-                    resolve(__dirname, `./logs/${logFile}`) :
-                    resolve(__dirname, `./logs/papiea_${logLevel}.log`) }),
-                new winston.transports.Console()
+                new winston.transports.File({
+                    filename: logFile ?
+                        resolve(__dirname, `./logs/${ logFile }`) :
+                        resolve(__dirname, `./logs/papiea_${ logLevel }.log`),
+                    format: winston.format.json()
```
nkntnx: Why not use the winstonFormat here too ?

nkntnx:
Can you also add a sample log file output to the review comments somewhere to see the new output ?

papiea-engine/src/logger.ts
Emulebest: I believe that we shouldn't confuse error logs with incorrect level logs. This way this might bloat the the error log with unexpected logs

papiea-engine/src/logger.ts
Emulebest: Pretty print insert a lot of indents which negatively affects log file size

papiea-engine/src/logger.ts
shlomi-nx: @Emulebest I agree with @nkntnx here. Logging something is better than silently ignore problems. I rather see bloated logs that point me to the problem which I could then fix, than this opaque `Unsupported logging level ` error which tells me that something somewhere gone wrong. 

If we get an unsupported logging level, I would implicitly make it as a `warning` level, and write that `Unsupported logging level` was used, along with the original log message and potentially its location in the code.

papiea-engine/src/logger.ts
shlomi-nx: We should not pre-optimize for file size.

papiea-engine/src/logger.ts
nkntnx: Fixed this.

papiea-engine/src/utils/utils.ts
```diff
@@ -70,4 +70,12 @@ export function isEmpty(obj: any) {
             return false;
     }
     return true;
+}
+
+export function safeJSONParse(chunk: string) {
+    try {
+        return JSON.parse(chunk)
+    } catch (e) {
+        return undefined
```
shlomi-nx: Should we just silently ignore errors? Can't this hide real problems?

shlomi-nx:
@Emulebest @nkntnx If this is done, please merge

PR #288, opened 2019-08-23
Add authentication options for Mongo

litvaOo:


papiea-engine/src/main.ts
```diff
@@ -40,6 +42,8 @@ const publicAddr: string = process.env.PAPIEA_PUBLIC_URL || "http://localhost:30
 const oauth2RedirectUri: string = publicAddr + "/provider/auth/callback";
 const mongoHost = process.env.MONGO_HOST || 'mongo';
 const mongoPort = process.env.MONGO_PORT || '27017';
+const mongoUser = process.env.MONGO_USER || '';
```
nkntnx: Since MONGO_USER is checked for truth value before using, an empty string does not add any value. Why have a default value for this variable ?

PR #292, opened 2019-09-02
Api docs per provider

### LINKED ISSUE #204: We should have api-docs per provider
We should have api-docs page per provider, and one for papiea itself (provider, s2s etc)

This way each provider, which is a different api domain, would have its own swagger file.

litvaOo:
resolves #204


papiea-engine/src/api_docs/api_docs_generator.ts
```diff
@@ -660,51 +664,48 @@ export default class ApiDocsGenerator {
...
+            if (kind.entity_procedures) {
+                Object.values(kind.entity_procedures).forEach(procedure => {
+                    paths[`/services/${ provider.prefix }/${ provider.version }/${ kind.name }/{uuid}/procedure/${ procedure.name }`] = {
+                        "post": this.callEntityProcedure(provider, kind, procedure)
+                    };
+                    Object.assign(schemas, procedure.argument);
+                    Object.assign(schemas, procedure.result);
+                });
+            }
+            if (provider.procedures) {
```
nkntnx: Since the "provider.procedures" is not connected to the "provider.kinds", should be taken out of the forEach.

adolgarev: @litvaOo ^^^ ?

litvaOo: resolved

papiea-engine/src/api_docs/api_docs_routes.ts
```diff
@@ -1,16 +1,55 @@
...
-            res.json(result);
-        }).catch(next);
+export default function createAPIDocsRouter(urlPrefix: string, apiDocsGenerator: ApiDocsGenerator, providerDb: Provider_DB) {
+    const apiDocsRouter = Router();
+
+    apiDocsRouter.use('/', swaggerUi.serve);
+
+    apiDocsRouter.get('/', async (req: Request, res: Response, next: NextFunction) => {
+        const providers = await providerDb.list_providers();
+        res.send(pug.renderFile(resolve(__dirname, 'main_swagger.pug'), {providers: providers}));
```
adolgarev: 1. Do not want to introduce template engine just because of one page, I'd better do not serve dynamic HTML at all
2. Add a list provider API call to ProviderAPIRouter and Provider_API
3. It should return ONLY providers you have access to, use existing auth mechanism
4. Obviously apiDocsRouter.get('/') page should make it possible to provide Authentication information, it would be great if instead of providing token itself one can do login through oauth like nunet-ui does
5. Having said that I expect this to be a static HTML served by backend and a static JS which will be just another client to papiea and do a) authenticate, b) list providers, c) provide urls for each provider to its api docs
6. The last but not least lets do some styling, using swagger UI css makes sense for me
Screenshot_20190904_163442

papiea-engine/src/api_docs/api_docs_routes.ts
```diff
@@ -1,16 +1,55 @@
-import * as express from "express";
+import { Request, Response, NextFunction, Router } from "express";
+import * as swaggerUi from "swagger-ui-express";
```
adolgarev: 1. Why to change express-swaggerize-ui to swagger-ui-express?
2. Iff you want to change then remove express-swaggerize-ui from package.json
3. I'd prefer to explicitly specify what we are importing to omit conflicts and understand what is accessible below, please do not use '*'

papiea-engine/src/api_docs/main_swagger.pug
```diff
@@ -0,0 +1,12 @@
+doctype html
```
adolgarev: see my comments above

papiea-engine/src/api_docs/api_docs_routes.ts
```diff
@@ -1,16 +1,55 @@
...
 import ApiDocsGenerator from "./api_docs_generator";
-const swaggerUi = require("express-swaggerize-ui");
+import { Provider_DB } from "../databases/provider_db_interface";
+import * as admin_swagger from './admin_swagger.json';
```
adolgarev: I'd prefer to explicitly specify what we are importing to omit conflicts and understand what is accessible below, please do not use '*'

papiea-engine/src/api_docs/api_docs_routes.ts
```diff
@@ -1,16 +1,55 @@
...
+import { Provider } from "papiea-core";
+import * as pug from "pug";
```
adolgarev: I'd prefer to explicitly specify what we are importing to omit conflicts and understand what is accessible below, please do not use '*'

papiea-engine/tsconfig.json
```diff
@@ -12,7 +12,8 @@
       "*": [
         "src/@types/*"
       ]
-    }
+    },
+    "resolveJsonModule": true,
```
adolgarev: Lets just read JSON file without importing it, what do you think?

papiea-engine/src/api_docs/api_docs_routes.ts
adolgarev: @litvaOo BTW if we move auth token to cookie as nunet-ui we may remove Authentication from swagger pages

papiea-engine/src/api_docs/api_docs_routes.ts
litvaOo: 1)`swagger-ui-express` provided a bit more flexibility in our case
2-3) Fixed

papiea-engine/src/api_docs/api_docs_routes.ts
litvaOo: Fixed

papiea-engine/src/api_docs/api_docs_routes.ts
litvaOo: Fixed

papiea-engine/tsconfig.json
litvaOo: Fixed

papiea-engine/src/api_docs/api_docs_routes.ts
adolgarev: > swagger-ui-express provided a bit more flexibility in our case
which ones?

papiea-engine/src/api_docs/api_docs_routes.ts
shlomi-nx: @adolgarev Very good points, see my corresponding responses:
1. I agree about not over-complicating for a single page with template engines
2. I agree with the API point
3. I tend to agree with that notion, however it may not be too easy to implement. Think that different providers may have different settings for oauth, and different authz models. How did you think of achieving this?
4. Same point as 3
5. Do you mean that this should be its own internal papiea-provider? Very interesting idea! We will need some general oauth provider for that, or use built-in user management, which may be an entity kind in that provider. Very interesting!
6. Absolutely! Lets use swagger UI css as mentioned

papiea-engine/src/api_docs/api_docs_routes.ts
shlomi-nx: I fully agree with not using '*' when importing

papiea-engine/src/api_docs/api_docs_routes.ts
litvaOo: @adolgarev 
I have a question about 4 - how should we chose provider, against which user will be authenticated?

litvaOo: @adolgarev 
Also, about your last comment - I do not think, that setting token to cookie, as we do on nunet-ui, is a good idea.
For example, if developer wants to test his policy from 3 different users perspective, right now it can be done, just with setting correct token on swagger page. If we'll switch to using cookies, it would require more actions(either setting cookie via some extensions, or re-login for each user), and become less convenient

papiea-engine/src/api_docs/api_docs_routes.ts
litvaOo: @adolgarev 
`express-swagerize-ui` is just a simple endpoint with one string substition in HTML, and does not provide as many options
`swagger-express-ui` provides separate methods for creating middleware and generating HTML page, and it made creation of separate pages easier, in the future it also might be useful for adding new options to our swagger pages(custom css/js/favicons/siteTitles, etc.)

papiea-engine/src/api_docs/api_docs_routes.ts
adolgarev: @litvaOo ok

papiea-engine/src/api_docs/main_swagger.pug
litvaOo: Removed

papiea-engine/src/api_docs/api_docs_routes.ts
shlomi-nx: @adolgarev I am not sure I want to do this kind of workflow at the moment. It is very convenient for me to be able to put any key I desire (such as s2s key) without going through a proper login mechanism. Also there is no single login mechanism for all of papiea, it is provider dependent. I say for now lets stick to what we have and we could make this kind of improvement in the future.

papiea-engine/src/api_docs/main_swagger.pug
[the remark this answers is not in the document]
litvaOo: Removed

PR #295, opened 2019-09-04
Filter check input params

### LINKED ISSUE #278: Filter accepts wrong filter clause and returns all authorized items
Currently, filter endpoint does not accept the following input:
```javascript
{ 'spec.provider_region': ping.native.host_region }
```

This causes all the allowed elements to be returned back, which is wrong. The following is the right query:
```javascript
{ spec: {'provider_region': ping.native.host_region }}
```

There are two ways of solving this:
1. Accept the first option as well as the second option
2. Throw a 400 bad request on the first option

A third way, is to optionally accept `fiql` query format as well as mongo format.

adolgarev:
Resolves #278

### DIFF
--- a/papiea-engine/src/validator/express_validator.ts
+++ b/papiea-engine/src/validator/express_validator.ts
@@ -1,0 +1,31 @@
+import { NextFunction, Request, Response } from "express";
+import { BadRequestError } from "../errors/bad_request_error";
+
+interface RequestValidatorOptions {
+    allowed_query_params: String[],
+    allowed_body_params?: String[]
+}
+
+export function check_request(request_validator_options: RequestValidatorOptions) {
+    const allowed_query_params_set = new Set(request_validator_options.allowed_query_params);
+    const allowed_body_params_set = request_validator_options.allowed_body_params ? new Set(request_validator_options.allowed_body_params) : undefined;
+    return (req: Request, res: Response, next: NextFunction) => {
+        for (var req_query_param in req.query) {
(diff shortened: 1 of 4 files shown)


papiea-engine/src/validator/express_validator.ts
```diff
@@ -0,0 +1,31 @@
...
+export function check_request(request_validator_options: RequestValidatorOptions) {
+    const allowed_query_params_set = new Set(request_validator_options.allowed_query_params);
+    const allowed_body_params_set = request_validator_options.allowed_body_params ? new Set(request_validator_options.allowed_body_params) : undefined;
+    return (req: Request, res: Response, next: NextFunction) => {
+        for (var req_query_param in req.query) {
+            if (!allowed_query_params_set.has(req_query_param)) {
+                throw new BadRequestError(`Allowed query params ${Object.keys(request_validator_options.allowed_query_params)}`);
+            }
+        }
+        if (request_validator_options.allowed_body_params && allowed_body_params_set) {
```
nkntnx: if request_validator_options.allowed_body_params is false, then, by virtue of the setting code above, allowed_body_params_set would be undefined. Do we need to check for both conditions here ?

nkntnx: Also, if req.body has params, but the allowed_body_params_set is false, then there will be no check. Is that the intented use ? If yes, please document it in the function/RequestValidatorOptions header.

Issue #261, opened 2019-08-08
Provide a log level that logs only request endpoint, response status


Emulebest:
related to audit logging from #304

PR #301, opened 2019-09-12
Added typescript SDK logging & Overall logging improvements

### LINKED ISSUE #300: Typescript SDK doesn't feature logging capabilities

### LINKED ISSUE #304: Make a separate audit logger that logs request info

Emulebest:
resolves #300 
resolves #304


Emulebest:
@shlomi-nx @nkntnx 
> Two points:
> 
> 1. Why is the var name called loggerfactory? As a user writing the provider, why can't I just get a log object that is already initialized with the right parameters to simply use?
> 2. The log object is a part of the context, it should not be passed explicitly in the handler functions

We need LoggerFactory for 2 reasons:
1) We need to dynamically resolve path to log file
`path.resolve(".", './logs/${ logPath }')`
If we use a preconstructed Log object inside SDK context then `path.resolve(".")` would point to `node_modules/papiea-sdk`
2) We tag every log line with https://github.com/nutanix/papiea-js/blob/1a4677bb35571f6ce8d67deeadf25e38a1324462/papiea-sdk/src/provider_sdk/typescript_sdk_logging.ts#L20 (require.main.filename) which is resolved to a filename where the procedure is defined
```
{
  message: 'Sum result 10',
  level: 'info',
  label: '/code/papiea-engine/__tests__/logging_tests/provider_sdk_logging.test.ts - computeSum'
}
```
Otherwise it would be
```
{
  message: 'Sum result 10',
  level: 'info',
  label: '/code/papiea-engine/node_modules/papiea-sdk/typescript_sdk.ts - computeSum'
}
```

These could be solved in 3 different ways:
1) Have a config file for SDK with variables used during runtime (e.g. log directory; resembles the idea of https://www.tutorialspoint.com/log4j/log4j_configuration.htm but for general papiea configuration). This could be also simplified via CLI tools like `create-react-app` or `ng new` have for React and Angular respectively, which create valid project structure
2) Have a default for log directory like `/etc/papiea/logs` and don't depend on runtime resolved vars https://github.com/nutanix/papiea-js/blob/1a4677bb35571f6ce8d67deeadf25e38a1324462/papiea-sdk/src/provider_sdk/typescript_sdk_logging.ts#L20 (require.main.filename)
3) Have a `get_logger(log_path?: string, log_level?: string): Logger` in SDK context, so that ergonomics would look like
```typescript
async (ctx, input) => {
    const logger = ctx.get_logger()
    logger.info(`Sum result ${ input.a + input.b }`)
    return input.a + input.b
}
```

IMO it is good to go with the option №3 for now, but aim for the option №1 because it a more flexible and extendable approach for the future use

papiea-engine/src/logger.ts
```diff
@@ -1,8 +1,21 @@
...
 import { Format } from 'logform';
+import { Logger, PapieaLogLevels } from "./logger_interface"
+import { UserAuthInfoRequest } from "./auth/authn"
+
+const papieaLogLevels: PapieaLogLevels = {
+    emerg: 0,
+    alert: 1,
+    crit: 2,
+    error: 3,
+    audit: 4,
```
nkntnx: audit can never be turned off !!!

papiea-engine/src/logger.ts
```diff
@@ -17,31 +30,31 @@ export function getLoggingMiddleware(logger: Logger) {
...
                 'Method': req.method,
                 'URL': req.originalUrl,
                 'Headers': req.headers,
-                'Status code': res.statusCode,
-                'Response body': res.body,
-                'Time': new Date(),
+                'Status_code': res.statusCode,
+                'Response_body': res.body,
             };
             if (req.method !== "GET") {
```
nkntnx: should this also check for a truthy res.body ?

papiea-engine/src/logger.ts
Emulebest: This just specifies priority for the logs https://github.com/winstonjs/winston#logging-levels

shlomi-nx: @Emulebest @nkntnx In my opinion, audit should not be complected with logging. If we want to use winston as a transport for audit, then it should be a different logger all together `audit.log(...)`

papiea-engine/src/logger.ts
Emulebest: What is the purpose, if it is undefined then it will be replaced by `{}`

nkntnx:
resolves #261

PR #302, opened 2019-09-16
S2s keys entity engine

### LINKED ISSUE #202: S2S keys - use entity engine
Papiea should probably have a notion of "papiea level kinds", meaning kinds that their provider is papiea itslef.

1. Lets see how we could use our existing entity mechanism to have s2s keys as a papiea level kind
2. If we decide not to use the entity mechanism, we should still have the structure of the api calls enforced (Should have name, uuid, etc)

litvaOo:
resolves #202

### DIFF
--- a/papiea-engine/src/provider/provider_routes.ts
+++ b/papiea-engine/src/provider/provider_routes.ts
@@ -98,7 +98,7 @@ export default function createProviderAPIRouter(providerApi: Provider_API) {
     }));
 
     providerApiRouter.delete('/:prefix/:version/s2skey/:uuid', CheckNoQueryParams, asyncHandler(async (req, res) => {
-        await providerApi.inactivate_key(req.user, req.body.uuid);
+        await providerApi.inactivate_key(req.user, req.params.uuid);
         res.json("OK");
     }));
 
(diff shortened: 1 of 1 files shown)


adolgarev:
> @litvaOo Is this it? Is this uses the entity engine?

@shlomi-nx I do not see at the moment how we can add entity which does not belong to some provider. Here I asked @litvaOo to make API look the same. This is mostly for UI so we can use same PapieaUQDC for entities and s2s keys without additional UQDC just for keys, as well making API consistent will mak its usage easier for end users.
I'll test PapieaUQDC against s2s keys. If we decide later to use entity engine for this we wont change API only implementation.

shlomi-nx:
Yes, thats why I thought we needed to create an internal "papiea" provider. In any case, having the same exact api is fine for now. Lets make sure it all gets exposed through the swagger UI, so it will be easy to learn about the apis.

> > @litvaOo Is this it? Is this uses the entity engine?
> 
> @shlomi-nx I do not see at the moment how we can add entity which does not belong to some provider. Here I asked @litvaOo to make API look the same for entities and s2s keys. This is mostly for UI so we can use same PapieaUQDC for entities and s2s keys without additional UQDC just for keys, as well making API consistent will mak its usage easier for end users.
> I'll test PapieaUQDC against s2s keys. If we decide later to use entity engine for this we wont change API only implementation.

PR #305, opened 2019-09-19
Router error logging

### LINKED ISSUE #272: Provider does not show on its side when a procedure return value validation failed
The client received an error such as

```
{"error":{"code":400,"errors":[{"message":"Target property 'cluster_snapshot_uuid' is not in the model"}],"message":"Procedure invocation failed."}}
```

But on the provider side nothing was printed or alerted.

What should happen when validation fails ON A RETURN VALUE:
1. The status code needs to be 500, as this is NOT a bad request, it is an internal server error, server being the provider. If validation fails on the user input, it should be 400
2. Provider logs must complain about the error, printing out the procedure name and the offending parts that did not validate properly

nkntnx:
The error path for Papiea-Engine does not log the error before returning. Adding that.

Resolved #272

### DIFF
--- a/papiea-engine/src/main.ts
+++ b/papiea-engine/src/main.ts
@@ -77,8 +77,7 @@ async function setUpApplication(): Promise<express.Express> {
         const papieaError = PapieaErrorImpl.create(err);
         res.status(papieaError.status)
         res.json(papieaError.toResponse())
-        logger.error(papieaError)
-        logger.error(err.stack)
+        logger.error(papieaError, err.stack)
     });
     return app;
 }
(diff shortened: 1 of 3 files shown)


papiea-engine/src/main.ts
```diff
@@ -77,6 +77,8 @@ async function setUpApplication(): Promise<express.Express> {
         const papieaError = PapieaErrorImpl.create(err);
         res.status(papieaError.status)
         res.json(papieaError.toResponse())
+        logger.error(papieaError)
```
Emulebest: Can we make those a single error message (Error - stack trace)?

PR #308, opened 2019-10-03
Intentful Core

### LINKED ISSUE #306: Implementing Intentful-core
The overall workflow as as follows:

1. In the SDK, one registers an SFS in a similar way to declaring an entity procedure (https://github.com/nutanix/papiea-js/issues/307)
2. At provider load time, all SFSs are compiled in
3. When a spec change arrives to Papiea, it runs the change through the SFSs to find potential matches. If no diff found, finish.
4. It then orders the matches (https://github.com/nutanix/papiea-js/issues/42)
5. picks one, generate an Intentful Task for it and executes the its handler (https://nutanix.github.io/papiea-js/Papiea-design.html#h-Intentful-Tasks-117)
6. when execution is done (tracked through the intentful task), goto 3

### LINKED ISSUE #307: Add intentful declarations in SDK
Need to add the following implementation to the SDK:

https://github.com/nutanix/papiea-js/blob/7b17bcb506e65c8004d47a9236329ecf3928803f/papiea-sdk/src/provider_sdk/typescript_sdk_interface.ts#L23-L27

Emulebest:
resolves #307 
resolves #306 

Breaking changes:
1) update_entity_spec (PUT endpoint) returns `IntentfulTask | null` instead of `[Metadata,Spec]`


papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -414,6 +414,15 @@ export class Kind_Builder {
...
     }
 
+    sfs_signature(sfs_signature: string, rbac: any, procedural_signature: Procedural_Signature): Kind_Builder {
+        this.kind.intentful_signatures.push({
+            signature: sfs_signature,
+            compiled_signature: null,
+            procedural_signature: procedural_signature
+        })
+        return this
+    }
```
shlomi-nx: Whats this function for? Is this instead of `on` function?

Emulebest: Yes, it is instead of `on` function. Having an `on` function implies a generic event handler. We could go with `on_sfs` or variation to give the event more clarity

shlomi-nx: @Emulebest It is a generic event handler. Papiea's generic even handler is an event of any change to an object content. The sfs is simply what defines the pattern matching of the event class we are listening on. Lets call it "on" for now and change it later if we choose so. 

Also, please accept a handler function by its direct signature, like we have for procedures, so `handler: (ctx: ctx:IntentfulCtx, entity: any)`, not `Procedural_Signature`.

shlomi-nx: So in the sdk, you could write:
```
const test_provider = ProviderSdk.create_provider(url, admin_s2s, provider_host, provider_port)

test_provider.prefix('test').version('0.0.1')

const table = test_provider.new_kind(load_yaml('kinds/table.yaml'))
table.on("legs", {}, (ctx, entity)=>{

  // The new entity's is guaranteed to have a difference between spec and status of 'legs' when this handler runs
  if (entity.spec.legs < 4 && entity.status.legs >= 4) console.log("someone just broke me")

})
```

papiea-core/src/core.ts
```diff
@@ -114,7 +105,7 @@ export interface Kind {
 
     intentful_behaviour: IntentfulBehaviour;
     //// Intentful behavior
-    intentful_signatures: Map<SFS, Intentful_Signature>;
+    intentful_signatures: Intentful_Signature[];
```
shlomi-nx: Why did you change this from a map?

papiea-engine/src/intentful_core/intentful_strategy_interface.ts
```diff
@@ -1,13 +1,66 @@
...
+    setKind(kind: Kind): void {
+        this.kind = kind
+    }
+
+    setUser(user: UserAuthInfo) {
+        this.user = user
+    }
+
+    async delete(metadata: Metadata): Promise<void> {
+        await this.dispatch("__delete", { metadata })
```
shlomi-nx: There may be a return value here that conditions the deletion process. Perhaps an exception handler or something.

papiea-core/src/core.ts
Emulebest: Intentful Signature interface already contains the connection between SFS and a handler, don't need to overcomplicate it here IMO

papiea-engine/src/intentful_core/intentful_strategy_interface.ts
Emulebest: Okay, got you, will do

papiea-engine/src/databases/task_db_interface.ts
```diff
@@ -0,0 +1,21 @@
+import { SortParams } from "../entity/entity_api_impl"
+import { Task } from "../tasks/task_interface"
+import { Provider, Kind } from "papiea-core"
+
+export interface Task_DB {
```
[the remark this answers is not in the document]
shlomi-nx: @Emulebest 👆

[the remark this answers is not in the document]
Emulebest: So, if I understood correctly, we create an Intentful Task first, send it to the invoking user and then assign a real task that resolves its diff and references an Intentful One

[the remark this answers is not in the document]
shlomi-nx: Almost. We first create an Intentful Task, but we do not 'assign' it to a real task - since there is no such thing. We MAY add async procedures which will have tasks, but thats not for now.

papiea-engine/src/databases/intentful_task_db_interface.ts
```diff
@@ -0,0 +1,21 @@
+import { SortParams } from "../entity/entity_api_impl"
+import { IntentfulTask } from "../tasks/task_interface"
+import { Provider, Kind } from "papiea-core"
+
+export interface IntentfulTask_DB {
+
+    create_task(task: IntentfulTask): Promise<void>
```
shlomi-nx: This name is confusing. Perhaps `insert_task` is more suitable?

papiea-engine/src/tasks/task_interface.ts
```diff
@@ -1,23 +1,29 @@
...
 
 // The task is started by a dedicated scheduler
-export interface Task {
+export interface IntentfulTask {
 
     // Identifier by which provider can change status of the task & user can monitor the execution
     uuid: uuid4
 
+    // Entity being modified by a task
+    entity: Entity
```
shlomi-nx: This should probably be an Entity Reference

papiea-engine/src/tasks/task_interface.ts
```diff
@@ -1,23 +1,29 @@
...
+
     // Diff resolved by this task
     diff: Diff
 
     // provider handler url, serves as an identifier for a type of task being executed
     handler_url: string
 
     // Running, Idle, Error, etc.
+    // TODO: Shlomi please design a state machine for this
```
shlomi-nx: An Intentful Task is never "Running". I see it as:
1. `Pending` - a spec change or a scheduled diffing created this intentful Task, it is not yet active
2. `Active` - meaning this Intentful Task is currently waiting for a diff to be resolved
3. `Completed Successfully` - All fields were set to the spec value at some point after the spec change was issued
4. `Completed Partially` - Some fields were set to the spec value, and some were not due to a newer spec version
5. `Failed` - none of the fields was changed to the given spec values, and there is already a newer spec version

papiea-engine/src/tasks/task_manager_interface.ts
```diff
@@ -1,54 +1,53 @@
...
+import { Entity } from "papiea-core"
+import Queue from "mnemonist/queue"
+import { Task } from "../../build/tasks/task_interface"
+import axios from "axios"
+
+class IntentfulTaskSet extends Set<IntentfulTask> {
+    get(task: Task): IntentfulTask | null {
+        for (let item of this) {
+            if (item.uuid === task.uuid) {
+                return item
```
shlomi-nx: Its a set, why the linear search?

papiea-engine/src/tasks/task_manager_interface.ts
```diff
@@ -1,54 +1,53 @@
...
-    has(task: Task) {
-        return this.taskMap.has(task.uuid)
+import { IntentfulTask_DB_Mongo } from "../databases/intentful_task_db_mongo"
+import { Entity } from "papiea-core"
+import Queue from "mnemonist/queue"
+import { Task } from "../../build/tasks/task_interface"
+import axios from "axios"
+
+class IntentfulTaskSet extends Set<IntentfulTask> {
+    get(task: Task): IntentfulTask | null {
```
shlomi-nx: What is this `Task`?

papiea-engine/src/tasks/task_manager_interface.ts
Emulebest: V8 set `has` is indeed O(1), however it doesn't have a `get` operation that returns a reference to the item in a set. General comment here is that Data Structures in JS and some of its libraries really lack in maturity and don't offer all the operations that are needed for our purposes

papiea-engine/src/tasks/task_interface.ts
Emulebest: I think this might be an `Entity` to avoid extra db queries, thus allowing the following snippet
```
                await axios.post(task.handler_url, {
                    metadata: task.entity.metadata,
                    spec: task.entity.spec,
                    status: task.entity.status,
                    input: task.diff.diff_fields
                })
```

papiea-engine/src/tasks/task_manager_interface.ts
Emulebest: Just a leftover from previous naming, will fix

papiea-engine/src/databases/intentful_task_db_interface.ts
Emulebest: `save_task`?

papiea-engine/src/tasks/task_manager_interface.ts
shlomi-nx: Can we use a proper DS then?

papiea-engine/src/tasks/task_interface.ts
shlomi-nx: That is not a good idea - this way you can have an old copy of the Entity which does not reflect newer changes. 

Please don't optimize before its needed. If we were to optimize this, it would probable be some kind of outside caching layer (think redis).

papiea-engine/src/databases/intentful_task_db_interface.ts
shlomi-nx: Sure, even though databases usually refer to `saving` as `inserting` ;)

papiea-engine/src/tasks/task_interface.ts
shlomi-nx: That is only useful if we want Intentful Task to capture the current version of the entity.

papiea-engine/src/tasks/task_manager_interface.ts
Emulebest: We might need to roll our own, I will take a look at https://www.collectionsjs.com/, it might come of use

papiea-engine/src/tasks/task_interface.ts
Emulebest: Okay, got it, will change to reference then

papiea-engine/src/main.ts
```diff
@@ -32,7 +33,8 @@ declare var process: {
         DEBUG_LEVEL: string,
         PAPIEA_ADMIN_S2S_KEY: string,
         LOGGING_LEVEL: string
-        PAPIEA_DEBUG: string
+        PAPIEA_DEBUG: string,
+        REDIS_URL: string
```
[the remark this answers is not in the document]
Emulebest: Okay, will remove it. Then we need another collection for the watchlist in Mongo, am I right?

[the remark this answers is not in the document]
Emulebest: Thinking about it, I feel like we need to introduce the interface that would allow us to switch seamlessly between mongo and redis (if we need that in the future). I am working on it now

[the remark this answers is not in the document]
shlomiv: @Emulebest Yes, you are right, we need a representation of the watchlist in mongo. Lets focus only on mongo (through an interface, as we have done), and worry about future optimizations later.

papiea-engine/src/tasks/differ_resolver.ts
```diff
@@ -48,18 +48,16 @@ export class DifferResolver {
...
+    public async onTask(task: IntentfulTask) {
         const tasks = this.watchlist.filter(entityTasks => entityTasks.entity_id === task.entity_ref.uuid)
         if (tasks.length === 0) {
             this.watchlist.push({
                 entity_id: task.entity_ref.uuid,
                 tasks: [task]
             })
             task.status = IntentfulStatus.Active
+            await this.intentfulTaskDb.update_task(task.uuid, task)
+            // TODO: start the handler and assign
```
Emulebest: @shlomi-nx Do we pass `diff_fields` as params to handler?

papiea-engine/src/tasks/differ_resolver.ts
```diff
@@ -68,7 +66,30 @@ export class DifferResolver {
...
+            const activeTasks = tasks[0].tasks.filter(task => task.status === IntentfulStatus.Active)
+            if (activeTasks.length === 0) {
+                return
+            } else if (activeTasks.length > 1) {
+                // TODO: This should be configurable via strategy
+                throw new Error(`There can not be more than one task running for entity with uuid: ${entity.uuid}`)
+            }
+            const activeTask = activeTasks[0]
+            if (metadata.spec_version >= activeTask.spec_version) {
+                activeTask.status = IntentfulStatus.Outdated
```
Emulebest: @shlomi-nx Do we mark this a `Outdated`, we don't have it as a part of State machine `IntentfulStatus`, or do we calculate the diff and mark it either `Failed` or `Completed Partially`?

papiea-engine/src/tasks/differ_resolver.ts
shlomi-nx: They should be in the context.

papiea-engine/src/tasks/differ_resolver.ts
shlomi-nx: This is only outdated if status is different from spec, and the new spec is different from the old spec, and the spec_version has increased. 

The outdated is the status of the sfs field we listen on.

The task's status is derived from the collection of all sfs fields:
1. Completed - all sfs fields are resolved successfully
2. Partially completed - all sfs fields are resolved, some successfully, some are outdated
3. Outdated - all sfs fields are resolved, all of which are outdated

papiea-engine/src/tasks/differ_resolver.ts
```diff
@@ -39,33 +39,67 @@ export class DifferResolver {
...
+                const [_, entityStatus] = await this.statusDb.get_status(activeTask.entity_ref)
+                for (let diff of activeTask.diffs) {
+                    // TODO: diff fields should go into the context
+                    await axios.post(diff.intentful_signature.procedural_signature.procedure_callback, {
+                        metadata: metadata,
+                        spec: entitySpec,
+                        status: entityStatus,
+                        input: diff.diff_fields
+                    })
+                    // TODO: find the handler id that is executing the procedure
```
Emulebest: How do we do that? Separate axios call to the handler?

papiea-engine/src/tasks/differ_resolver.ts
```diff
@@ -39,33 +39,67 @@ export class DifferResolver {
...
+                const [metadata, entitySpec] = await this.specDb.get_spec(activeTask.entity_ref)
+                const [_, entityStatus] = await this.statusDb.get_status(activeTask.entity_ref)
+                for (let diff of activeTask.diffs) {
+                    // TODO: diff fields should go into the context
+                    await axios.post(diff.intentful_signature.procedural_signature.procedure_callback, {
+                        metadata: metadata,
+                        spec: entitySpec,
+                        status: entityStatus,
+                        input: diff.diff_fields
+                    })
```
Emulebest: There is a question about from the point of what user are we executing the procedure?

shlomi-nx: The behavior is exactly the same as regular procedures. Starts out with the identity of the provider, and if required, in code, the token will be switched to that of the user (see create cluster procedure for example)

papiea-engine/src/tasks/differ_resolver.ts
shlomi-nx: Thats the mechanism we discussed for verifying that the handler is still running. Not sure I get your question.

papiea-core/src/core.ts
```diff
@@ -76,30 +98,22 @@ export interface Entity_Reference  {
...
+    Differ = "differ"
 }
 
 // The differ is used to locate a diff in an entity between the
 // current status and the desired state. 
 export interface Differ {
 
     // Get the next diff from an entity based on the 
-    next_diff(entity:Entity_Reference, spec: Spec, status: Status):Diff;
+    diffs(kind: Kind, spec: Spec, status: Status): Generator<Diff, any, undefined>;
```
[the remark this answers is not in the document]
Emulebest: `diffs` has a better ergonomics as a generator IMO. You don't need to put to in a explicit loop to iterate through each individual diff. For the list representation we've got `all_diffs(kind: Kind, spec: Spec, status: Status): Diff[]`

papiea-core/src/core.ts
```diff
@@ -60,7 +82,7 @@ export type SFS = string;
 export interface Intentful_Signature {
     signature: SFS;
     compiled_signature: any
-    function_callback: Provider_Callback_URL;
+    procedural_signature: Procedural_Signature;
```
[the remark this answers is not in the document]
Emulebest: I think you are right, will do

papiea-core/src/core.ts
```diff
@@ -16,6 +16,28 @@ export type Version = string;
...
+    Completed_Successfully = 2,
+
+    // Some fields were set to the spec value, and some were not due to a newer spec version
+    Completed_Partially = 3,
+
+    // None of the fields was changed to the given spec values,
+    // and there is already a newer spec version
+    Failed = 4,
+
+    Outdated = 5,
```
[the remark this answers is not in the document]
Emulebest: Failed is a temporary state signifying that we should retry the spec change, Outdated is triggered when we rewrite the spec with a new version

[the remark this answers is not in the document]
shlomi-nx: We never rewrite the spec with a new version. The way I was thinking about this, there is no purpose in Failed. As long as it is not outdated or done it needs to be retried. Outdated should mean that we are stopping to retry because the desire has already changed.

Does this makes sense?

papiea-core/src/core.ts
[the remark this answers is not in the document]
shlomi-nx: If this was in the engine I'd feel more comfortable with this, but since this is in core, which will be used by the sdk, it should be relevant to all languages, this needs to be something that is not language specific. I much rather have a trivial "get next" implementation, and a generator could be implemented in the individual sdks layers, each targeting a particular language.

Generators are essentially lazy lists, or streams. There is no purpose in making this lazy, since the asker wants to know the state of the pending diffs at the moment of invocation, not at time of processing. The purpose of "next_diff" was to establish some ordering of this list, based on the logic inside Papiea. Getting all diffs allows the provider to apply its own logic into picking up which diff to work on next, having a generator does not help with that at all. 

Given this, diffs should return a plain list of all diffs that represents the list of diffs at the moment of invocation (on the sdk we can expose this through generators if it makes sense). next_diff should return the next diff the engine thinks should be handled at the time of invoking it.

Does this makes sense?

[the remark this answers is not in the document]
Emulebest: I think you are right about next_diff being more verbose to show the fact that we might do internal ordering. I wasn't expecting any kind of optimization when introducing iterators based on generators. I thought of it more like of a code in convenient, idiomatic way. Sure, we can simplify the definition so that it looks more universal across different core definitions

papiea-core/src/core.ts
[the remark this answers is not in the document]
Emulebest: So essentially `Completed_Partially` should be retried, if so, than I agree with that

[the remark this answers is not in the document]
shlomi-nx: Hmm.. My original intention with `Completed_Partially` is that some fields managed to settle to the specified spec (the completed part), but some got outdated (the partial part). So, the difference between `Completed_Partially` and `Outdated` is that the first have at least one field that managed to change to the desired spec, while the second has none.

PR #315, opened 2019-10-21
Add predownloaded cert

### LINKED ISSUE #314: Add possibility to mount pre-downloaded rds-combined-ca-bundle.pem
Sometimes during starting up papiea with secured connection to MongoDB/Document DB, attempt to download `rds-combined-ca-bundle.pem` fails with `EAI_AGAIN`.
To avoid it, we need to add possibility to mount this certificate and avoid downloading it on each start.

litvaOo:
Resolves #314

### DIFF
--- a/papiea-engine/src/databases/mongo.ts
+++ b/papiea-engine/src/databases/mongo.ts
@@ -1,4 +1,5 @@
 import axios from "axios";
+import { promisify } from "util"
 import { MongoClient, Db } from "mongodb";
 import { Spec_DB_Mongo } from "./spec_db_mongo";
 import { Status_DB_Mongo } from "./status_db_mongo";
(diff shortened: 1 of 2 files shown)


papiea-engine/src/databases/mongo.ts
```diff
@@ -37,6 +37,9 @@ export class MongoConnection {
     }
 
     async download_rds_cert(): Promise<void> {
+        if (fs.existsSync('rds-combined-ca-bundle.pem')) {
+            return;
+        }
```
adolgarev: please do not use *Sync, use in this case https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original

litvaOo: Done

PR #320, opened 2019-10-25
Added utc timestamp to sdk

### LINKED ISSUE #318: Make UTC timestamps mandatory in all logs
Conformant canonical time stamps look like this: 2019-10-21 16:36:19Z. Note that ISO 8601 permits other variations, for example reducing or increasing accuracy, which may make sense in some use cases. The trailing Z is mandatory for UTC time stamps.

Emulebest:
resolves #318


Emulebest:
@adolgarev the example would be 
timestamp: '2019-10-25T08:42:02.769Z', so it is UTC according to ISO 8601

PR #325, opened 2019-11-13
Update spec security fix

### LINKED ISSUE #326: Update spec security issue
If one passes fake owner in extension it is being used in casbin. Instead read extension from db.

adolgarev:
Resolves #326


papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -114,11 +114,14 @@ export class Entity_API_Impl implements Entity_API {
         const provider = await this.get_provider(prefix, version);
         const kind = this.find_kind(provider, kind_name);
         this.validate_spec(spec_description, kind, provider.allowExtraProps);
-        const metadata: Metadata = { uuid: uuid, kind: kind.name, spec_version: spec_version, extension: extension } as Metadata;
+        const entity_ref: Entity_Reference = { kind: kind_name, uuid: uuid };
+        const metadata: Metadata = (await this.spec_db.get_spec(entity_ref))[0];
         await this.authorizer.checkPermission(user, { "metadata": metadata }, Action.Update);
+        metadata.spec_version = spec_version;
+        metadata.extension = extension;
```
Emulebest: Why do we reassign the extension? `strategy.update(metadata, spec_description)` doesn't update metadata in any way in PUT endpoint

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -114,11 +114,14 @@ export class Entity_API_Impl implements Entity_API {
...
         const strategy = this.intentfulCtx.getIntentfulStrategy(kind.intentful_behaviour)
-        const [_, spec] = await strategy.update(metadata, spec_description)
-        return [metadata, spec];
+        const [res_metadata, spec] = await strategy.update(metadata, spec_description)
```
Emulebest: I can't see metadata changed in this workflow

papiea-engine/src/entity/entity_api_impl.ts
adolgarev: it doesn't update extension, but it updates spec_version

papiea-engine/src/entity/entity_api_impl.ts
adolgarev: spec_version

PR #328, opened 2020-03-12
Fix to differ strategy automatically assigning status to new spec

Emulebest:


shlomi-nx:
Ok, found the problem that caused the differ to think the diff is resolved - spec was returned instead of status.

The next item to solve is why the diff handler is called a few times in a row? It should be invoked once per iteration.

shlomi-nx:
The test `Change single field intentful workflow should pass` is brittle at the moment, if it does not pass a "retry" on the job may have it succeed. Need to investigate why.

shlomi-nx:
To test just this:
```bash
docker-compose exec papiea-engine node_modules/.bin/jest  -t 'Change single field intentful workflow should pass'
```

PR #338, opened 2020-03-24
Fix to intentful task being deleted too fast

### LINKED ISSUE #335: Completed intentful tasks are deleted immediately
Completed tasks should be kept around for users to be able to look them up.

Perhaps they should not be deleted at all (perhaps some cleanup job could remove really old ones, or archive them).

Emulebest:
resolves #335

### DIFF
--- a/papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
+++ b/papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
@@ -76,25 +76,25 @@ describe("Intentful Workflow tests", () => {
             let retries = 4
             try {
             for (let i = 1; i <= retries; i++) {
-                    const result = await entityApi.get(`/intentful_task/${task.uuid}`)
-
-                    // SHLOMI: I think that the task is getting removed too quickly once it is completed
-                    // task is returning 
-                    // console.debug("WHAT IS GOING ON??!", result.data, i)
+                const result = await entityApi.get(`/intentful_task/${task.uuid}`)
                 if (result.data.status === IntentfulStatus.Completed_Successfully) {
                     expect(result.data.status).toEqual(IntentfulStatus.Completed_Successfully)
-                    return
+                    await timeout(120000)
+                    try {
+                        // Wait till task is deleted
+                        await entityApi.get(`/intentful_task/${task.uuid}`)
+                    } catch (e) {
+                        // TODO: This should be 404, I believe
+                        expect(e.response.statusCode).toBe(500)
+                        return
+                    }
                 }
                 await timeout(7000)
             }
             } catch (e) {
-                console.debug("Got some error:", e.response.status)
+                console.debug(e.response)
+                // console.debug("Got some error:", e.response.status)
             }
-
-            // Shlomi: In any case, this should succeed if the diff got resolved
-            const { data: entity } = await entityApi.get(`/${sdk.provider.prefix}/${sdk.provider.version}/${kind_name}/${metadata.uuid}`)
-            expect(entity.spec).toEqual(entity.status)
-
         } finally {
             sdk.server.close();
         }
--- a/papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
+++ b/papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
@@ -75,17 +75,18 @@ describe("Intentful Workflow tests", () => {
             })
             let retries = 5
             try {
-            for (let i = 1; i <= retries; i++) {
-                const result = await entityApi.get(`/intentful_task/${task.uuid}`)
-                if (result.data.status === IntentfulStatus.Completed_Successfully && result.data.marked_for_deletion) {
-                    expect(result.data.status).toEqual(IntentfulStatus.Completed_Successfully)
-                    expect(result.data.marked_for_deletion).toBeDefined()
-                    return
+                for (let i = 1; i <= retries; i++) {
+                    const result = await entityApi.get(`/intentful_task/${ task.uuid }`)
+                    if (result.data.status === IntentfulStatus.Completed_Successfully) {
+                        expect(result.data.status).toEqual(IntentfulStatus.Completed_Successfully)
+                        break
+                    }
+                    await timeout(7000)
                 }
-                await timeout(7000)
-            }
+                const res = await entityApi.get(`/${ sdk.provider.prefix }/${ sdk.provider.version }/${ kind_name }/${ metadata.uuid }`)
+                expect(res.data.status.x).toEqual(20)
             } catch (e) {
-                console.log(e.response)
+                console.log(`Couldn't get intentful task: ${e.response}`)
             }
         } finally {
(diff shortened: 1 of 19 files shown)


papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
```diff
@@ -73,28 +73,20 @@ describe("Intentful Workflow tests", () => {
...
-                    const result = await entityApi.get(`/intentful_task/${task.uuid}`)
-
-                    // SHLOMI: I think that the task is getting removed too quickly once it is completed
-                    // task is returning 
-                    // console.debug("WHAT IS GOING ON??!", result.data, i)
-                if (result.data.status === IntentfulStatus.Completed_Successfully) {
+                const result = await entityApi.get(`/intentful_task/${task.uuid}`)
+                if (result.data.status === IntentfulStatus.Completed_Successfully && result.data.marked_for_deletion) {
                     expect(result.data.status).toEqual(IntentfulStatus.Completed_Successfully)
+                    expect(result.data.marked_for_deletion).toBeDefined()
```
shlomi-nx: Rethinking about this, why do we need an explicit `marked_for_deletion'? 

When should a task be deleted? When its in a state that cannot make any progress, or a terminal state,  such as `Active`, `Outdated` or `Failed`.

Isn't a `last_status_change_time` a better approach? We could then delete all tasks that are in a terminal state for a minimum amount of time?

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
```diff
@@ -73,28 +73,20 @@ describe("Intentful Workflow tests", () => {
...
                     return
                 }
                 await timeout(7000)
             }
             } catch (e) {
-                console.debug("Got some error:", e.response.status)
+                console.log(e.response)
```
shlomi-nx: Please include some text. Searching for this later will be impossible.

papiea-engine/src/databases/intentful_task_db_mongo.ts
```diff
@@ -14,19 +14,25 @@ interface TaskAggregation {
...
-        async init(): Promise<void> {
+    async init(): Promise<void> {
         try {
             await this.collection.createIndex(
                 { "uuid": 1 },
                 { unique: true },
             )
+            await this.collection.createIndex(
+                { "marked_for_deletion": 1 },
+                { expireAfterSeconds: this.deletedTaskPersistSeconds }
```
shlomi-nx: Even though this is nice, I don't think its appropriate here. We may want to reconfigure this during runtime, or may even want to pause task deletion altogether. Putting this data in the database's internal makes us lose control.

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
Emulebest: A task is marked for deletion when is in `if (task.status !== IntentfulStatus.Active && task.status !== IntentfulStatus.Pending && task.status !== IntentfulStatus.Failed)`. It seems kind of similar to last_status_change_time

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
Emulebest: Sure

papiea-engine/src/databases/intentful_task_db_mongo.ts
Emulebest: Maybe we can tie this to the concrete task in the runtime

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
shlomi-nx: Exactly, so I think we should feel comfortable renaming this field to reflect its true meaning. Then we could use another cleanup process (not mongo's ttl) to remove the outdated tasks.

papiea-engine/src/databases/intentful_task_db_mongo.ts
shlomi-nx: I am not sure I understand what you mean by that

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
Emulebest: Since we are already running differ as a background process, we might want to do cleanup there. What do you think?

papiea-engine/src/databases/intentful_task_db_mongo.ts
Emulebest: I mean make the deletion time not global but a field on intentful task structure to make it dynamic

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
shlomi-nx: Sure, that sounds appropriate.

Emulebest:
Since this is going more and more into design space, I'd like to leave my complete overview on how I see this:
Preconditions:
1) Add optional `default_completed_task_expiry` field to entity
2) Add optional `completed_task_expiry` field to intentful_task
3) Add null default `completed_at` field to intentful_task
4) Add user endpoint  PUT `/intentful_task/:id` for updating expiry (maybe narrow down it to PUT `/intentful_task/:id/expiry`) with permissions required for updating the entity that intentful task is bound to

Workflow:
1) Entity A is created with additional `default_completed_task_expiry` field set to 120 seconds (default: 60 seconds)
2) Task B is created with `completed_task_expiry` set to `default_completed_task_expiry` of the entity A
3) optional: User updates the expiry to be 180 seconds by the task id
4) Task reaches `Completed_Successfully` or `Completed_Partially` state, `completed_at` field is set to `new Date()`
5) `clearFinishedTasks()` checks that `completed_at + completed_task_expiry < new Date()`, when the condition is met, meaning task is ready to be deleted, it is deleted from the database

@shlomi-nx Please make your comments

shlomi-nx:
I don't understand why we need all this complexity. Users has no say when a task is deleted from DB. DB cleanup is a process entirely in the engine control. It should probably be in days and not seconds.

Very simple: store the last status change timestamp. A cleanup process physically deletes it if its in a terminal state for a duration exceeding some maximum. 

We can take this over a call if you prefer, but lets not over-complicate such a simple manner. As far as I am concerned, we could simply not cleanup now.

Emulebest:
> I don't understand why we need all this complexity. Users has no say when a task is deleted from DB. DB cleanup is a process entirely in the engine control. It should probably be in days and not seconds.
> 
> Very simple: store the last status change timestamp. A cleanup process physically deletes it if its in a terminal state for a duration exceeding some maximum.
> 
> We can take this over a call if you prefer, but lets not over-complicate such a simple manner. As far as I am concerned, we could simply not cleanup now.

Okay, I thought that we were aiming for more runtime configurable manner. Less is more though, I will make it the easy way

PR #340, opened 2020-03-24
Python SDK

### LINKED ISSUE #336: Create python papiea_sdk
It should have same abilities as the typescript sdk.

adolgarev:
Resolves #336


papiea-python-sdk/papiea/python_sdk.py
```diff
@@ -0,0 +1,292 @@
...
+        return res
+
+    async def get(self, prefix: str, headers: dict = {}) -> str:
+        new_headers = {}
+        new_headers.update(self.headers)
+        new_headers.update(headers)
+        async with session.get(self.base_url + "/" + prefix,
+                                headers=new_headers) as resp:
+            res = await resp.text()
+        return res
```
Emulebest: If the `session.get` fails what do we return?

papiea-python-sdk/papiea/python_sdk.py
```diff
@@ -0,0 +1,292 @@
...
+    async def post(self, prefix: str, data: dict, headers: dict = {}) -> str:
+        new_headers = {}
+        new_headers.update(self.headers)
+        new_headers.update(headers)
+        data_binary = json.dumps(data).encode("utf-8")
+        async with session.post(self.base_url + "/" + prefix,
+                                data=data_binary,
+                                headers=new_headers) as resp:
+            res = await resp.text()
+        return res
```
Emulebest: If the `session.post` fails what do we return?

papiea-python-sdk/papiea/example.py
```diff
@@ -0,0 +1,199 @@
...
+    async with ClientSession() as session:
+        async with session.put(
+            ctx.url_for(entity),
+            data=data_binary,
+            headers={
+                "Content-Type": "application/json",
+                "Authorization": f"Bearer {requesting_user_token}",
+            },
+        ) as resp:
+            res = await resp.text()
```
shlomi-nx: This portion should be extracted to use a papiea client

papiea-python-sdk/papiea/example.py
```diff
@@ -0,0 +1,199 @@
...
+            f"{location_entity_base_url}",
+            data=data_binary,
+            headers={
+                "Content-Type": "application/json",
+                "Authorization": f"Bearer {user_s2s_key}",
+            },
+        ) as resp:
+            res = await resp.text()
+            logger.debug(f"Created entity {res}")
+            entity_uuid = json.loads(res)["metadata"]["uuid"]
```
shlomi-nx: Also this, without a client this is very verbose.

papiea-python-sdk/papiea/example/example.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+        await sdk.register()
+        server = sdk.server
+
+        user_s2s_key = await create_user_s2s_key(sdk)
+        async with EntityCRUD(
+            PAPIEA_URL, PROVIDER_PREFIX, PROVIDER_VERSION, "Location", user_s2s_key
+        ) as entity_client:
+            entity = await entity_client.create(
+                Spec(x=10, y=11),
```
shlomiv: (I am not familiar with python, so these may be silly questions)

Is it mandatory to use this syntax? How would users use it for embedded structures (e.g. maps with vectors of maps, etc..)?

Is `{"x":10,"y":11}` also supported?

papiea-python-sdk/papiea/example/example.py
```diff
@@ -0,0 +1,176 @@
...
+            procedure_move_output_definition,
+            move_x,
+        )
+        await sdk.register()
+        server = sdk.server
+
+        user_s2s_key = await create_user_s2s_key(sdk)
+        async with EntityCRUD(
+            PAPIEA_URL, PROVIDER_PREFIX, PROVIDER_VERSION, "Location", user_s2s_key
+        ) as entity_client:
```
shlomiv: (I am not familiar with python, so these may be silly questions)

Two questions:
1. Why is `EntityCRUD` constructor behind an async call? It does no IO? I would expect using it to use async, not defining it..
2. Is this syntax common in modern python? Why not simply use await:
     ```
     entity_client = await EntituCRUD(...)
     entity = await entity_client.create(...)
     ```

papiea-python-sdk/papiea/example/example.py
```diff
@@ -0,0 +1,176 @@
...
+                Spec(x=10, y=11),
+                metadata_extension={
+                    "owner": "alice",
+                    "tenant_uuid": "ada14b27-c147-4aca-9b9f-7762f1f48426",
+                },
+            )
+            logger.debug(f"Created entity {entity}")
+            entity_uuid = entity.metadata.uuid
+            res = await entity_client.invoke_procedure(
+                "moveX", EntityReference(uuid=entity_uuid), 5
```
shlomiv: In TypeScript's sdk, the entity's metadata is considered its Reference, so this would simply be:
```
res = await entity_client.invoke_procedure("moveX", entity.metadata, 5)
```

Why do we need this verbosity in Python?

papiea-python-sdk/papiea/example/example.py
adolgarev: > Is {"x":10,"y":11} also supported?
yes, one can pass any dict here

papiea-python-sdk/papiea/example/example.py
adolgarev: This is a async context manager. See https://docs.python.org/3/reference/datamodel.html#context-managers for context managers and https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with for async ones. In short this syntax is used to automatically clean up resources when async clean up is needed

papiea-python-sdk/papiea/example/example.py
adolgarev: this should work too, let me write example this way

papiea-python-sdk/papiea/python_sdk_context.py
```diff
@@ -0,0 +1,101 @@
...
+            entity_reference.kind,
+            self.get_invoking_token(),
+        )
+
+    async def check_permission(
+        self,
+        entity_action: List[Tuple[Action, EntityReference]],
+        provider_prefix: Optional[str] = None,
+        provider_version: Optional[Version] = None,
+    ) -> bool:
```
shlomi-nx, line 41: This needs an optional user-token. If not given the default should be the invoking user.

adolgarev: done

PR #341, opened 2020-03-25
On_(create|delete) proceed() workflow

### LINKED ISSUE #339: `on_(create|delete)` - should not receive SFS. Improve tests
1. `on_(create|delete)` have no use of an SFS
2. The handlers are to be invoked synchronously for the user. So a `POST` on an entity would only respond when the handler is done running, either with the created object, or with the thrown error message 
2. Throwing an error inside the handler functions should not proceed with the execution
3. Should add tests for throwing errors inside these handlers

Emulebest:
resolves #339

### DIFF
--- a/papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
+++ b/papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
@@ -1022,4 +1022,83 @@ describe("SDK callback tests", () => {
             sdk.server.close();
         }
     });
+
+    test("On delete callback with error should interrupt execution", async () => {
+        expect.hasAssertions();
+        const sdk = ProviderSdk.create_provider(papieaUrl, adminKey, server_config.host, server_config.port);
+        const location = sdk.new_kind(location_yaml);
+        const prefix = "provider_on_delete_callback"
+        sdk.version(provider_version);
+        sdk.prefix(prefix);
+        location.on_delete({}, Procedural_Execution_Strategy.Halt_Intentful, {}, {}, async (ctx, input) => {
+            throw new Error("Cannot invoke on delete")
+        })
+        try {
+            await sdk.register()
+            kind_name = sdk.provider.kinds[0].name
+            const { data: { metadata } } = await entityApi.post(`/${ prefix }/${ provider_version }/${ kind_name }`, {
+                spec: {
+                    x: 10,
+                    y: 11
+                }
+            }, {
+                headers: {
+                    'Authorization': `Bearer ${ adminKey }`
+                }
+            })
+            try {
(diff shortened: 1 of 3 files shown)


papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -1022,4 +1022,83 @@ describe("SDK callback tests", () => {
...
+            })
+            try {
+                await entityApi.delete(`/${prefix}/${provider_version}/${kind_name}/${metadata.uuid}`, {
+                    headers: {
+                        'Authorization': `Bearer ${adminKey}`
+                    }
+                })
+            } catch (e) {
+                expect(e.response.data).toBeDefined()
+                expect(e.response.data.error.message).toBe("Error: Dispatch couldn't be called, Error occurred.")
```
shlomi-nx: The error message should be the text in the error `Cannot invoke on delete`.

In any case, this error message is misleading - dispatch is the process of finding the function to call - the function was found and called, its execution got terminated due to a legitimate cause.

papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -1022,4 +1022,83 @@ describe("SDK callback tests", () => {
...
+                        y: 11
+                    }
+                }, {
+                    headers: {
+                        'Authorization': `Bearer ${adminKey}`
+                    }
+                })
+            } catch (e) {
+                expect(e.response.data).toBeDefined()
+                expect(e.response.data.error.message).toBe("Error: Dispatch couldn't be called, Error occurred.")
```
shlomi-nx: Same comment as `on_delete`

papiea-sdk/src/provider_sdk/typescript_sdk_context_impl.ts
```diff
@@ -73,12 +73,63 @@ export class ProceduralCtx implements ProceduralCtx_Interface {
...
+        let token: string | null = null
+        try {
+            token = this.get_invoking_token()
+        } catch (e) {
+        }
+        try {
+            if (token !== null) {
+                await axios.delete(this.url_for(this.entity as Entity), { headers: { 'Authorization': `Bearer ${ token }` } });
+            } else {
+                await axios.delete(this.url_for(this.entity as Entity));
```
shlomi-nx: I am not certain that proceed should not be its own rest endpoint that is listening for a particular handler (could be implemented using some guid), similar to is_alive task endpoint

shlomi-nx:
If this PR does not handle `proceed`, lets name it appropriately

Emulebest:
This is very dated and needs to be redesigned in the future

PR #342, opened 2020-03-26
On_(create|delete) Improved error propagation & error messages

### LINKED ISSUE #339: `on_(create|delete)` - should not receive SFS. Improve tests

Emulebest:
resolves #339

### DIFF
--- a/papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
+++ b/papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
@@ -1053,7 +1053,7 @@ describe("SDK callback tests", () => {
                 })
             } catch (e) {
                 expect(e.response.data).toBeDefined()
-                expect(e.response.data.error.message).toBe("Error: On Delete couldn't be called; Cannot invoke on delete.")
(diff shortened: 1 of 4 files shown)


papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -1022,4 +1022,82 @@ describe("SDK callback tests", () => {
...
+                        y: 11
+                    }
+                }, {
+                    headers: {
+                        'Authorization': `Bearer ${adminKey}`
+                    }
+                })
+            } catch (e) {
+                expect(e.response.data).toBeDefined()
+                expect(e.response.data.error.message).toBe("Error: On Create couldn't be called; Cannot invoke on create.")
```
shlomi-nx: Lets have the string identical to the thrown string. We could add some more context to the error if absolutely required, but I dont see a reason to prefix with `Error:`

PR #343, opened 2020-03-27
Intentful engine cleanups

Emulebest:


Emulebest:
Deprecating this in favor of #356

PR #354, opened 2020-03-31
Added task creation from changing status

### LINKED ISSUE #347: `ctx.update_status` seems to not add an entity to the watch list
for diff resolution.

Emulebest:
resolves #347


Emulebest:
Deprecating this in favor of #356

PR #356, opened 2020-04-03
Entity Watchlist & Intentful Tasks separated

### LINKED ISSUE #347: `ctx.update_status` seems to not add an entity to the watch list

### LINKED ISSUE #359: Handle a list of intentful kinds
See discussion in https://github.com/nutanix/papiea-js/pull/356.
TLDR; now it is quite hard to get a list of random intentful entities for example, you'd need to: query all providers where at least on kind is intentful, find intentful kinds in the queried providers, collect them in a list, select random entities where entity.name in selected kind list.
This issue is aimed at resolving this inconvenience

Emulebest:
resolves #347 
resolves #359


papiea-engine/src/tasks/diff_resolver.ts
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+            // Check if we have already rediffed this entity
+            if (diffs === undefined) {
+                [diffs, metadata, spec, status] = await this.rediff(entry_reference)
+            }
+            if (diffs.length === 0) {
+                this.removeDiff(entry_reference)
+                return
+            }
```
shlomiv, line 104: This looks like a potential race condition, since diffs may be added between the check and removing from the watchlist.

For this reason, we should have the system randomly pick a batch of entities at every iteration and add them to the watchlist. The diff resolution will get delayed based on the ratio between batch size and total entity size.

papiea-engine/src/tasks/diff_resolver.ts
```diff
@@ -0,0 +1,111 @@
...
+            // No active diffs found, choosing the next one
+            const [executing_diff, execution_delay] = await this.launchOperation(diffs[0], metadata!, spec!, status!)
```
shlomiv, line 106: We should extract selecting the diff to work on to a function we could later change based on some strategy. 

1. Trivial strategy, next diff (what you have)
2. Deepest strategy, diffs that resolves deeper sfs's will get executed first
3. Some kind of an "intelligent" diff selector, we could discuss on this later

papiea-engine/src/tasks/diff_resolver.ts
```diff
@@ -0,0 +1,111 @@
...
+        await axios.post(diff.intentful_signature.procedure_callback, {
+            metadata: metadata,
+            spec: spec,
+            status: status,
+            input: diff.diff_fields
+        })
+        const delay = {
+            delay_seconds: 20,
+            delaySetTime: new Date()
+        }
```
shlomiv, line 69: The intenetful handler may or may not return a delay value. If it does not return, we should populate with the default, which should be configurable per kind.

papiea-engine/src/tasks/diff_resolver.ts
Emulebest: Totally agree, was going to mark in the comments actually

papiea-engine/src/tasks/diff_resolver.ts
Emulebest: That was also intended, I made this static only for now, later it is going to be as we discussed

papiea-engine/src/tasks/diff_resolver.ts
shlomi-nx: We could turn this into its own ticket if you would like. However, I would at least select one at random before pushing this.

papiea-engine/src/tasks/diff_resolver.ts
Emulebest: This is one solution, I'll also look into inter-process locking that might suit our situation

shlomi-nx: This may not be useful, as papiea will run in a cluster, and different instances may modify the watchlist. I rather use some optimistic concurrency to locking.

Another potential mechanism would be to have a "ref-counted" watchlist.

Emulebest: Do we want the batch size to be static (e.g. env variable) or dynamic (e.g. amount of entities/10)

shlomi-nx: For now lets have a configurable constant. We may change the strategy later.

papiea-engine/src/databases/spec_db_mongo.ts
```diff
@@ -122,6 +124,44 @@ export class Spec_DB_Mongo implements Spec_DB {
         });
     }
 
+    async list_random_specs(size: number, fields_map?: any, sortParams?: SortParams): Promise<([Metadata, Spec])[]> {
+        const filter: any = {};
+        if (fields_map) {
+            filter[ "metadata.deleted_at" ] = datestringToFilter(fields_map.metadata.deleted_at)
```
shlomi-nx, line 130: whats the purpose of this? There is no purpose at selecting deleted entities

papiea-engine/src/tasks/diff_resolver.ts
```diff
@@ -0,0 +1,181 @@
...
+            if (diffs === undefined) {
+                try {
+                    [diffs, metadata, kind, spec, status] = await this.rediff(entry_reference)
+                } catch (e) {
+                    this.logger.debug(`Couldn't rediff entity with uuid ${entry_reference.entity_reference.uuid}: ${e}. Removing from watchlist`)
+                    await this.removeDiff(entry_reference)
+                    continue
+                }
+            }
+            if (diffs.length === 0 || kind!.intentful_behaviour !== IntentfulBehaviour.Differ) {
```
shlomi-nx, line 131: Why would entities that that have `kind!.intentful_behaviour !== IntentfulBehaviour.Differ` even enter the watchlist?

papiea-engine/src/tasks/diff_resolver.ts
```diff
@@ -0,0 +1,181 @@
...
+        }
+    }
+
+    private static async calculate_batch_size(): Promise<number> {
+        return 5
+    }
+
+    // This method is needed to avoid race condition
+    // when diffs may be added between the check and removing from the watchlist
+    // the batch size maybe static or dynamic
```
shlomi-nx, line 162: Did not quite get this part

papiea-engine/src/tasks/diff_resolver.ts
```diff
@@ -0,0 +1,181 @@
...
+    private async populateWatchlist() {
```
shlomi-nx, line 163: Perhaps a better name would be `addRandomEntitiesToWatchlist`.

papiea-engine/src/tasks/diff_resolver.ts
Emulebest: `list_random_specs` is used to get random entities to enter watchlist. Since getting ONLY intentful entities is a cumbersome and resource intensive thing to do because of our db design (query all providers where at least on kind is intentful, find intentful kinds in the queried providers, collect them in a list, select random entities where entity.name in selected kind list). I assume that in the real world scenario we will mostly have intentful entities and it would be easier to remove spec-only and basic ones if they get randomly chosen in a list

papiea-engine/src/tasks/diff_resolver.ts
Emulebest: This is just the explanation of why we need to add random entities at selected interval

papiea-engine/src/tasks/diff_resolver.ts
Emulebest: I don't like long names for simple function but I'll design a better name :)

papiea-engine/src/databases/spec_db_mongo.ts
Emulebest: See the explanation below. will remove the `metadata.deleted_at` part, just copied the whole function from `list_spec` so didn't notice that

papiea-engine/src/tasks/diff_resolver.ts
Emulebest: Alright, got your idea

papiea-engine/src/tasks/diff_resolver.ts
shlomi-nx: Hmm.. I don't particularly like this workaround. We should think of ways of indexing all the Kinds that are intentful, and select only non-deleted entities items from these kinds.

Also, this logic fits better in `addRandomEntitiesToWatchlist` - only valid entities should be added in the first place.

Emulebest: Agreed, I'll move it to `addRandomEntitiesToWatchlist` for now and open a new issue to come up with a better solution

papiea-engine/src/tasks/intentful_listener_mongo.ts
```diff
@@ -1,96 +1,88 @@
...
+                }
+            }
+            if (this.watchlist.has(entry_reference)) {
+                const status_entry = this.statuses.get(entry_reference)
+                if (!status_entry || JSON.stringify(status_entry) !== JSON.stringify(status)) {
+                    await this.onSpec.call(entry_reference, status)
+                }
+            }
+        }
     }
```
shlomi-nx, line 61: I don't understand this. 

You are making a linear scan on all entities in the db twice, and then comparing if the spec/status is same as itself in quick succession? This hardly seems correct, do you want to talk about this?

We essentially know when there is a spec/status call, because the api goes through us. Never directly to the DB.

Emulebest: You are right, I forgot to add a comment that this is only for now to keep it similar to the resolver working already in master. I know that https://developer.mongodb.com/quickstart/nodejs-change-streams-triggers might be a fit for us, but since it is a stream it means some shenanigans with deserialization and getting familiar with this implementation might arise. I am more of `make it work, make it optimal` approach so didn't succumb to doing it in this commit but a great notice, thanks.
Answering your other question, yes we know the spec change at the API call but since the API serving and diff resolution happen in two different nodejs processes we would need some form of IPC between those and I don't really like the native functions nodejs gives us, so it is either watching their shared piece (db) or IPC of some form (pub/sub)

shlomi-nx: Ok, great. Just wanted to make sure we are on the same page. Please make a ticket to handle that. 

I never intended to use any kind if direct IPC, I like strong decoupling. This could be implemented in a similar fashion to the watchlist on the DB or some redis. 

nodejs-change-streams-triggers seems interesting. Will it work with other DBs? Lets try to make this simple.

papiea-engine/src/intentful_core/intentful_strategies/differ_intentful_strategy.ts
```diff
@@ -39,6 +42,20 @@ export class DifferIntentfulStrategy extends IntentfulStrategy {
             task.diffs.push(diff)
         }
         await this.intentfulTaskDb.save_task(task)
+        const watchlist = await this.watchlistDb.get_watchlist()
```
shlomi-nx, line 45: This function creates a task with the current `spec_version`, saves the task, saves the spec change and returns the task.

The way that I see this, the task needs to be created without a `spec_version`, spec change should attempt to CAS, if it was successful the `spec_version` updates to the one it was CASsed into, and the task is returned. If CAS failed, the task should turn to `failed` status with an appropriate message, and returned.

papiea-engine/src/tasks/task_resolver.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    private async onIntentfulHandlerRestart(entity: EntryReference) {
+        const tasks = await this.intentfulTaskDb.list_tasks({ entity_ref: entity.entity_reference })
+        const failed_task = tasks.find(task => task.status === IntentfulStatus.Failed)
+        if (failed_task) {
+            failed_task.status = IntentfulStatus.Active
+            await this.intentfulTaskDb.update_task(failed_task.uuid, { status: failed_task.status })
+        }
+    }
```
shlomi-nx, line 139: Oh I see, this will turn failed tasks to active again.. Since a task is an object for communicating with the user, `Failed` status usually means a terminal state. I would not use that as a mechanism for retrying, as it may confuse watching users.

papiea-engine/src/databases/intentful_task_db_mongo.ts
```diff
@@ -34,6 +34,7 @@ export class IntentfulTask_DB_Mongo implements IntentfulTask_DB {
 
     async save_task(task: IntentfulTask): Promise<void> {
         task.created_at = new Date()
+        task.spec_version += 1
```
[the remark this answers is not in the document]
shlomi-nx: Please see comments below, particularly https://github.com/nutanix/papiea-js/pull/356#discussion_r408400686

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
```diff
@@ -39,13 +39,13 @@ describe("Intentful Workflow tests", () => {
...
 
-    test("Change single field intentful workflow should pass", async () => {
+    test("Change single field differ resolver should pass", async () => {
         expect.assertions(2);
         const sdk = ProviderSdk.create_provider(papieaUrl, adminKey, server_config.host, server_config.port);
         try {
             const location = sdk.new_kind(locationDataDescription);
             sdk.version(provider_version);
-            sdk.prefix("location_provider_intentful");
+            sdk.prefix("location_provider_intentful_1");
```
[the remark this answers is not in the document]
Emulebest: Yes, but anyways, that's a new kind name, so no collisions occur. Just wanted a distinction in terms of different providers, that makes it easier for debugging. I'll make the cleanup here anyways, good notice

papiea-engine/docker-compose-debug.yml
```diff
@@ -18,4 +18,5 @@ services:
                         - HOT_RELOAD=true
                         - PAPIEA_DEBUG=true
         mongo:
-                image: mongo:4.1
\ No newline at end of file
+                image: mongo:4.1
+                command: mongod --replSet rs0
```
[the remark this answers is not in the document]
Emulebest: Mongo streams don't work without mongo being a replica set

papiea-engine/src/databases/mongo.ts
```diff
@@ -61,13 +65,31 @@ export class MongoConnection {
         }
     }
 
-    async connect(): Promise<void> {
+    async connect(master: boolean): Promise<void> {
```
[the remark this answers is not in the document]
Emulebest: Now as we run mongo in replica set mode, this mode needs to be initialized. As we have 2 processes, an attempt to init the replica set mode twice results in an error. That's why I call the API process master, as it tries to init the mode and the diff resolver process not a master , thus it gets only the configuation of replica set mode

papiea-engine/src/tasks/intentful_listener_mongo_stream.ts
```diff
@@ -0,0 +1,79 @@
...
+                const entry_reference: EntryReference = {
+                    provider_reference: {
+                        provider_prefix: entity.metadata.provider_prefix,
+                        provider_version: entity.metadata.provider_version
+                    },
+                    entity_reference: {
+                        uuid: entity.metadata.uuid,
+                        kind: entity.metadata.kind
+                    }
+                }
```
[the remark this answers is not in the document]
Emulebest: I will try if this is created from the same `entity` everywhere, not to have a function with a lot of optional params like (entity?, metadata?, entity_reference?)

papiea-engine/src/tasks/intentful_listener_mongo_stream.ts
```diff
@@ -0,0 +1,79 @@
...
+            console.error(e)
+            throw e
+        }
+    }
+
+    private async watchCollection() {
+        this.changeStreamIterator = this.entityDbCollection.watch([
+            { $match: { operationType: "update" } }
+        ], { fullDocument: 'updateLookup', resumeAfter: this.resumeToken })
+        this.changeStreamIterator.on("change", async (next) => {
```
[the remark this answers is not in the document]
Emulebest: You might be right

papiea-engine/src/tasks/intentful_listener_mongo_stream.ts
```diff
@@ -0,0 +1,79 @@
...
+                        await this.onStatus.call(entry_reference, entity.status)
+                        return
+                    }
+                    for (let key of Object.keys(next.updateDescription.updatedFields)) {
+                        const partial_status = key.split(".")
+                        if (partial_status[ 0 ] === "status") {
+                            await this.onStatus.call(entry_reference, entity.status)
+                            return
+                        }
+                    }
```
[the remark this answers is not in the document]
Emulebest: Alright, this is a bit confusing, I should move it to a function. The logic is that some updates get in form of `updatedFields: { status: [Object] }` when we replace the status, but remember we also have the partial updates so when performing `PATCH` request we get `updatedFields: { 'status.x': [Object] }`, thus need the loop logic to check both variants. The important thing is that they don't overlap

papiea-engine/src/tasks/task_resolver.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    private async rediff(entry_reference: EntryReference, metadata: Metadata, spec: Spec, status: Status): Promise<Diff[]> {
+        const provider = await this.providerDb.get_provider(entry_reference.provider_reference.provider_prefix, entry_reference.provider_reference.provider_version)
+        const kind = this.providerDb.find_kind(provider, metadata.kind)
+        return this.differ.all_diffs(kind, spec, status)
+    }
+
+    private async onSpec(entity: EntryReference, specVersion: number, spec: Spec) {
```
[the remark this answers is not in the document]
Emulebest: Okay, the ways I implemented it:
1) OnSpec is called when a new spec is received and a new task is created. We set the new to task to Active and move the states of tasks that have just become outdated. We rediff those and see if they are `Completed Successfully`, `Completed Partially` or `Outdated` by this point.
2) OnStatus is called when a new status is received and a task presumably already exists. It works only with a single `Active` task. It can move it only to `Completed Successfully`. Then is takes a new Pending task and makes it Active

See this as following:
A task is considered `Completed Successfully` when all the diffs were resolved (by moving status to spec) and the task
A task could get to `Completed Partially` and to `Outdated` only be spec updates, which means that the `Completed Partially` has some diffs resolved but is not relevant (spec version older) and the `Outdated` has no diffs resolved and is irrelevant (spec version older)

papiea-engine/src/tasks/task_resolver.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    private async onIntentfulHandlerFail(entity: EntryReference) {
+        const tasks = await this.intentfulTaskDb.list_tasks({ entity_ref: entity.entity_reference })
+        const active_task = tasks.find(task => task.status === IntentfulStatus.Active)
+        if (active_task) {
+            active_task.status = IntentfulStatus.Failed
+            await this.intentfulTaskDb.update_task(active_task.uuid, { status: active_task.status })
+        }
+    }
```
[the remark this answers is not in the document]
Emulebest: Let's definitely talk about this

papiea-engine/src/tasks/task_resolver.ts
Emulebest: Let's talk about this semantic

papiea-engine/__tests__/intentful_engine_tests/intentful_workflow.test.ts
```diff
@@ -76,17 +77,63 @@ describe("Intentful Workflow tests", () => {
...
-                        break
+                    const res = await entityApi.get(`/intentful_task/${ task.uuid }`)
+                    if (res.data.status === IntentfulStatus.Completed_Successfully) {
+                        expect(res.data.status).toBe(IntentfulStatus.Completed_Successfully)
+                        const result = await entityApi.get(`/${ sdk.provider.prefix }/${ sdk.provider.version }/${ kind_name }/${ metadata.uuid }`)
+                        expect(result.data.status.x).toEqual(20)
+                        return
                     }
-                    await timeout(7000)
+                    await timeout(10000)
```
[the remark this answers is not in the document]
Emulebest: I could explain the problem on the call

papiea-engine/src/databases/intentful_task_db_mongo.ts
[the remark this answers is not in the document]
Emulebest: Yes, this is CAS case, will move it to constructor (initially it was in the constructor, moved it here because of JS reasons :) )

papiea-engine/docker-compose-debug.yml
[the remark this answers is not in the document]
shlomi-nx: I see. I am not at ease with using mongo streams for these watches. I rather us make a general purpose solution that would work regardless of the DB used.

papiea-engine/src/databases/mongo.ts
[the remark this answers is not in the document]
shlomi-nx: So now we need to handle multiple process that cannot be identical. I do not like it.

papiea-engine/src/tasks/task_resolver.ts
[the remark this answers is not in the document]
shlomi-nx: Regardless, this info could be deduce by `spec`, `status` and `spec_version`. The context of coming from a spec change or status change should not be a relevant. 

We would know that a change came from a spec change if `spec_version` has increased.

.circleci/docker-compose.yml
```diff
@@ -10,4 +10,5 @@ services:
                 depends_on:
                         - mongo
         mongo:
-                image: mongo:4.1
\ No newline at end of file
+                image: mongo:4.1
+                command: /bin/bash -c 'mongod --replSet rs0 --bind_ip_all & pid=$$! && sleep 5 && mongo --eval "rs.initiate()" && wait $$pid'
```
shlomi-nx, line 14: Do we still need this? Aren't we not using mongo streams anymore?

Emulebest: We are but we can switch from streams to simple polling at any time

Issue #358, opened 2020-04-07
SDK does not log unhandled exceptions


joshua-berry-ntnx:
If an unhandled exception occurs during execution of a Papiea procedure or intent handler, a 500 error is sent back to the Papiea engine, with some (but not all?) error details included in a blob of JSON.

The unhandled exception should _also_ be logged to stderr in a human-readable format (and possibly a machine-readable format if running in production mode).

I would consider this issue to be high-priority—I've already lost a couple hours of debugging time to this, because I didn't realize the SDK-caught errors weren't being logged, and because the actual exception was getting swallowed and replaced with a generic "Error while making request" exception somewhere along the way (I still need to look into how/why the original error got swallowed).

joshua-berry-ntnx:
![](https://github.trello.services/images/mini-trello-icon.png) [Address Papiea issues needed for inventory-only BMaaS](https://trello.com/c/V6gmTU73/66-address-papiea-issues-needed-for-inventory-only-bmaas)

Emulebest:
This is already resolved in a PR linked, waiting for moving logging to core

joshua-berry-ntnx:
Closed by #363

Issue #357, opened 2020-04-07
SDK and engine log output is very difficult/impossible for humans to parse


joshua-berry-ntnx:
Here is an example from BMaaS. This is a log message for which the relevant bit of information is that a new entity state is being applied:

```
{"entity":{"metadata":{"kind":"machine","spec_version":2,"uuid":"54f6b695-cfe4-4670-a23e-548323012f37","created_at":"2020-04-07T22:25:11.302Z","extension":{"owner":"me"}},"spec":{"state":"online"},"status":{"state":"starting"}},"diff":[{"keys":{},"key":"state","spec-val":["online"],"status-val":["starting"]}],"level":"info","message":"Entity diff for on_state","timestamp":"2020-04-07T22:25:14.340Z","label":"/Users/joshua.berry/src/BMaaS-POC/dist/main.js - state"}
```

Now imagine digging through 15, 20, or hundreds of these to find the one you're looking for.

I understand the need to generate machine-readable logs for monitoring/alerting purposes, but when running in non-production mode, can we at least pretty-print the extra parameters, and put the actual log message at the beginning rather than buried somewhere in the middle of the JSON object?

joshua-berry-ntnx:
![](https://github.trello.services/images/mini-trello-icon.png) [Address Papiea issues needed for inventory-only BMaaS](https://trello.com/c/V6gmTU73/66-address-papiea-issues-needed-for-inventory-only-bmaas)

Emulebest:
This is already resolved in a PR linked, waiting for moving logging to core

PR #363, opened 2020-04-20
papiea-sdk: Improve logging format when not in production [#357, #358]

joshua-berry-ntnx:
- Provide nicer/human-readable log messages when not running in
  production, so it's easier to debug provider (and SDK) issues.

- Spit out error logs when an unhandled exception bubbles up to the SDK
  from within a procedure or intent handler.


joshua-berry-ntnx:
Partially addresses #357 (more to come once early feedback has been received)

joshua-berry-ntnx:
[Address Papiea issues needed for inventory-only BMaaS](https://trello.com/c/V6gmTU73/66-address-papiea-issues-needed-for-inventory-only-bmaas)

shlomi-nx:
@joshua-berry-ntnx I see tests are failing, lets make sure we fix the tests before pushing

joshua-berry-ntnx:
@shlomi-nx Yes absolutely. This is not ready to push yet, I just wanted something to start the discussion, especially around where the logger functionality should go (since I think we can probably leverage this in the engine as well).

(BTW, I am seeing engine tests fail in master so I'm not sure if my changes are actually causing the test failures. Will investigate further.)

shlomi-nx:
Thanks for the clarification. Oh! Master is breaking!?? I was assuming it was fine.. Thanks for verifying.

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -545,4 +548,20 @@ export class Kind_Builder {
         return this
     }
 }
-export {Version, Kind, Procedural_Signature, Provider, Data_Description, Procedural_Execution_Strategy, Entity, ProceduralCtx_Interface, Provider_Power, IntentfulCtx_Interface, UserInfo, S2S_Key}
\ No newline at end of file
+
+function make_context(provider: ProviderSdk,
```
Emulebest, line 552: Can we move it to be a static function for either Provider SDK class or Context class? It seems a bit unbound atm

Emulebest:
> Partially addresses #357 (more to come once early feedback has been received)

Also addresses #358

joshua-berry-ntnx:
> > Partially addresses #357 (more to come once early feedback has been received)
>
> Also addresses #358

Haha oops, yes it does.

papiea-engine/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -490,6 +490,36 @@ describe("Provider Sdk tests", () => {
             sdk.server.close();
         }
     });
+
+    test("Provider with provider level procedures should correctly handle exceptions in the provider", async () => {
```
joshua-berry-ntnx, line 494: Nice, thanks for adding the test :)

papiea-sdk/src/provider_sdk/typescript_sdk.ts
joshua-berry-ntnx: Looking at it a bit more, I think we can actually just fold it into the ProceduralCtx constructor itself, which now only seems to be used by make_context().

joshua-berry-ntnx:
How do you guys feel about moving the logging stuff into papiea-core and using it in both -engine and -sdk? And should we do that here or start a separate PR?

shlomi-nx:
Sounds good. I would do it here, it should be rather small.

package.json
```diff
@@ -5,10 +5,10 @@
...
-    "build-all": "npm --prefix papiea-core run build &&  npm --prefix papiea-engine run build && npm --prefix papiea-client run build && npm --prefix papiea-sdk run build",
-    "patch-all": "npm --prefix papiea-core version patch &&  npm --prefix papiea-engine version patch && npm --prefix papiea-client version patch && npm --prefix papiea-sdk version patch",
-    "install-all": "npm --prefix papiea-core install &&  npm --prefix papiea-engine install && npm --prefix papiea-client install && npm --prefix papiea-sdk install",
+    "clean-all": "for p in papiea-core papiea-backend papiea-engine papiea-sdk papiea-client; do npm --prefix $p run clean || exit 1; done",
+    "build-all": "for p in papiea-core papiea-backend papiea-engine papiea-sdk papiea-client; do npm --prefix $p run build || exit 1; done",
+    "patch-all": "for p in papiea-core papiea-backend papiea-engine papiea-sdk papiea-client; do npm --prefix $p version patch || exit 1; done",
+    "install-all": "for p in papiea-core papiea-backend papiea-engine papiea-sdk papiea-client; do npm --prefix $p install || exit 1; done",
```
shlomi-nx, line 11: Show off! :D

papiea-backend/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "papiea-backend",
```
shlomi-nx, line 2: Straight up "backend" may be misleading. Something like "backend-utils" may be more prescriptive. WDYT?

package.json
joshua-berry-ntnx: Would I ever do that? 😎

papiea-backend/package.json
joshua-berry-ntnx: Good point. I was struggling to come up with a good name but you're right, "backend" implies it's the _whole_ back end. I'll see if I can think of something better.

papiea-backend-utils/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "papiea-backend",
```
shlomi-nx, line 2: Forgot to rename here

papiea-backend-utils/src/logging.ts
```diff
@@ -0,0 +1,197 @@
...
+import {Format} from 'logform'
+
+import {inspect} from 'util'
+
+export const LOG_LEVELS = {
+    emerg: 0,
+    alert: 1,
+    crit: 2,
+    error: 3,
+    audit: 4,
```
shlomi-nx, line 11: Are we certain that "audit" is a log level?

We could think that audit itself may have various levels.

joshua-berry-ntnx: I don't think it is, however the -engine code was written in this way before (it defined its own log levels, of which `audit` was one). Since this is mostly a refactoring job, I opted to keep it that way for now, especially since I didn't know the rationale for doing it this way to begin with. (I suspect it was to get an "audit" keyword in the log output in a standard place, just like we would get "info", "warning", etc.)

shlomiv: Fair enough. We should probably remove it from here at some point.

joshua-berry-ntnx:
This latest push should fix all the tests. If/when CircleCI passes this time around, I think this is ready to merge.

PR #369, opened 2020-04-23
Initial Benchmarks

### LINKED ISSUE #365: Initial performance benchmark
We need to start gathering some performance characteristics of Papiea. We need the following:

1. A simple, no-op provider that:
   1. has a three procedures: entity, kind and provider level
   2. has an intentful operation that does nothing but resolve the diff
2. Run it on a typical server machine in AWS
3. Create a client which blasts requests (can run few client instances in parallel)
   1. Measure DB operations only (create, filter, get, delete)
   2. Measure operations involving provider:
       1. Procedure invocations (each)
       2. Intentful invocation (PUT requests):
           1. Each testing client should work with his own entities, so no race conditions
           1. Measure two things: 
               a. Time to succeed setting the operation (CAS)
               b. Time to complete operation (Intentful handler done)
               c. As an extra step, we could have the intentful handler fail every so often to account for retries.

Emulebest:
resolves #365


papiea-engine/__benchmarks__/src/utils.ts
```diff
@@ -0,0 +1,17 @@
+import * as autocannon from "autocannon";
+
+export function autocannon_mutate_req(opts: autocannon.Options, initial_state: any, update_func: (state: any) => any) {
+    return new Promise((resolve, reject) => {
+        const instance = autocannon(opts, (err: Error, res: any) => {
+            resolve(res);
+        })
```
shlomiv, line 7: Where is this function used?

Emulebest: It is planned to be used tomorrow in intentful workflow benchmarks

PR #375, opened 2020-04-30
Benchmarks add Makefile & CF template

Emulebest:


Makefile
```diff
@@ -0,0 +1,55 @@
+run-papiea: benchmark-install
```
shlomiv, line 1: why is `benchmark-install` a prerequisite of `run-papiea`

Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+run-benchmark-local: run-papiea
+	cd ./papiea-engine; \
+	docker-compose exec papiea-engine bash -c 'cd __benchmarks__ && npm i && npm run bench-local'
+
+
+clean-all:
+	npm run clean-all; \
+	cd ./papiea-engine/__benchmarks__; \
+	rm -rf node_modules
```
shlomiv, line 55: Should this remove all node_modules?

shlomiv:
@joshua-berry-ntnx, do you have comments on this makefile?

Makefile
```diff
@@ -0,0 +1,55 @@
...
+	npm install
+
+
+stop-papiea:
+		cd ./papiea-engine; \
+    	if [ -z `docker-compose ps -q papiea-engine` ] || [ -z `docker ps -q --no-trunc | grep $$(docker-compose ps -q papiea-engine)` ]; then \
+			echo 'Papiea not running'; \
+		else \
+		  docker-compose kill; \
+        fi
```
joshua-berry-ntnx, line 33: Seems like you have some whitespace problems in the rule above, at least when I view it in GitHub (which annoyingly doesn't show tab characters explicitly)

Makefile
```diff
@@ -0,0 +1,55 @@
+run-papiea: benchmark-install
+	cd ./papiea-engine; \
+	if [ -z `docker-compose ps -q papiea-engine` ] || [ -z `docker ps -q --no-trunc | grep $$(docker-compose ps -q papiea-engine)` ]; then \
+      docker-compose up -d; \
+	  for i in `seq 1 10`; \
+	  do \
+		docker-compose logs --tail=5 papiea-engine | grep 'app listening on port' && echo Success && exit 0; \
+		sleep 15; \
```
joshua-berry-ntnx, line 8: Let's poll more frequently so we can start faster. If you are running these benchmarks repeatedly, the time can really add up...

Makefile
```diff
@@ -0,0 +1,55 @@
...
+		docker-compose logs --tail=5; \
+	  done \
+    fi
+
+
+papiea-build:
+	npm run build-all
```
joshua-berry-ntnx, line 15: Build should depend on install.

Makefile
```diff
@@ -0,0 +1,55 @@
...
+
+
+papiea-install:
+	npm run install-all
```
joshua-berry-ntnx, line 19: If you want to skip the install if it's not needed, I suggest replacing the `papiea-install` target with something like this:

```make
node_modules package-lock.json: package.json
	npm install
	touch node_modules package-lock.json

node_modules: package-lock.json
```

And then making `papiea-build` depend on `node_modules`.

joshua-berry-ntnx: (btw in looking at the bmaas code for an example, I realized I did this wrong there, haha)

Makefile
```diff
@@ -0,0 +1,55 @@
+.PHONY: run-papiea papiea-build stop-papiea run-tests run-benchmark run-benchmark-local
```
joshua-berry-ntnx, line 1: I don't know what make will do if `.PHONY` is the first target, but I suggest putting an explicit default target above this as the first line, something like `default: papiea-build`

Makefile
joshua-berry-ntnx: We should maybe have two separate targets, one which removes only direct build artifacts and one which removes node_modules. (In BMaaS I called mine `clean` and `distclean` because I was following historical automake precedent, but ¯\_(ツ)_/¯)

If you want a nice, generic `clean-all`/`distclean` rule, try this for removing everything in `.gitignore`: `git clean -dXf`

Makefile
Emulebest: It skips .PHONY, but I'll address the comment

Makefile
```diff
@@ -0,0 +1,67 @@
+default: run-benchmark-local
```
shlomi-nx, line 1: I would not make the running the benchmark the default target.

This is the Makefile for this project, usually the default target is to compile and build

papiea-engine/src/docker-compose-benchmark.yml
```diff
@@ -0,0 +1,18 @@
...
+                container_name: papiea-engine
+                build: .
+                image: nutanix-docker.jfrog.io/papiea:base
+                volumes:
+                        - ../:/code
+                depends_on:
+                        - mongo
+                ports:
+                        - "3333:3000"
+                environment:
```
shlomi-nx, line 13: Shouldn't this file be in `__benchmark__` folder?

PR #377, opened 2020-05-04
Changed Mongo poll implementations and fixed intersection bugs

Emulebest:


papiea-backend-utils/src/logging.ts
```diff
@@ -46,7 +46,6 @@ export class LoggerFactory {
     readonly options: LoggerOptions
 
     private static readonly INSPECT_OPTIONS = {
-        colors: true, depth: 3, maxStringLength: 120,
```
shlomi-nx, line 49: Why did we remove colors?

papiea-engine/src/tasks/intentful_listener_mongo_simple.ts
```diff
@@ -31,16 +29,15 @@ export class IntentfulListenerMongo implements IntentfulListener {
                 this.entities.set(metadata.uuid, [spec, status])
                 continue
             }
-            if (deepEqual(spec, entry[0]) || deepEqual(status, entry[1])) {
+            if (!deepEqual(spec, entry[0]) || !deepEqual(status, entry[1])) {
```
shlomi-nx, line 32: Not entirely sure this is needed.

PR #378, opened 2020-05-05
Added papiea errors to client

### LINKED ISSUE #376: Errors thrown by papiea-client should be well-defined
[This might be a dup of or related to #255.]

tl;dr: I would like papiea-client to catch and re-throw (or re-reject) axios failures with some types exposed through the papiea-client interface which are specific to the error that occurred.

Right now it looks like papiea-client simply allows axios exceptions to bubble up to the caller. The problem with this is the caller is probably in another module which depends on papiea-client, but may not depend on axios. So it's very difficult for callers to name the type of exception they are catching, e.g.:

```ts
try {
    // ... papiea-client call ...
} catch (e) {
    if (e instanceof ???) {
        ...
    } else { throw e; }
}
```

Similarly, there are several different kinds of errors thrown by papiea-client (and/or papiea-engine) in different situations. For example, update conflicts, poorly-formatted requests, authorization errors, etc. These errors should probably be distinguishable as Papiea errors rather than generic HTTP errors, network errors, etc., and it is often useful to distinguish between different kinds of Papiea errors (poorly-formatted spec, update conflict, etc.).

Some ideas/examples worth considering (I'm mainly focusing on `update()` because that's where this first occurred to me), feel free to ignore my suggested names:

- `InvalidSpecError` - entity spec validation failed in update() or create(), perhaps with additional details about what failed in the error message
- `UpdateConflictError` - update() with old/incorrect spec_version()
- `NotFoundError` - get() which returns no entity (or maybe this can just return undefined, if that's the current behavior)
- `NotAuthorizedError`

Emulebest:
resolves #376 

It seems like we are still going for papiea-core package to be a DTO so I decided not to move errors there. Moreover, another reason is having flexibility in processing errors on papiea and on the client. Also exceptions should not be a contract, they should be a convenience, JSON structure of an error should be a contract

### DIFF
--- a/papiea-client/src/entity_client.ts
+++ b/papiea-client/src/entity_client.ts
@@ -1,17 +1,54 @@
-import axios from "axios";
-import { Metadata, Spec, Entity_Reference, Entity, EntitySpec } from "papiea-core";
+import axios, { AxiosPromise, AxiosRequestConfig } from "axios";
+import { Metadata, Spec, Entity_Reference, Entity, EntitySpec, PapieaErrorTypes } from "papiea-core";
+import {
+    BadRequestError,
+    ConflictingEntityError,
+    EntityNotFoundError,
+    PermissionDeniedError,
+    ProcedureInvocationError,
+    PapieaServerError,
+    UnauthorizedError,
+    ValidationError
+} from "./errors/errors";
+
+function make_request<T = any>(f: (url: string, data?: any, config?: AxiosRequestConfig) => AxiosPromise<T>, url: string, data?: any, config?: AxiosRequestConfig): AxiosPromise<T> {
+    try {
+        return f(url, data, config)
+    } catch (e) {
+        switch (e?.response?.data?.error.type) {
+            case PapieaErrorTypes.ConflictingEntity:
+                throw new ConflictingEntityError(e.response.data.error.message)
+            case PapieaErrorTypes.PermissionDenied:
+                throw new PermissionDeniedError(e.response.data.error.message)
+            case PapieaErrorTypes.EntityNotFound:
+                throw new EntityNotFoundError(e.response.data.error.message)
+            case PapieaErrorTypes.ProcedureInvocation:
+                throw new ProcedureInvocationError(e.response.data.error.message)
+            case PapieaErrorTypes.Unauthorized:
+                throw new UnauthorizedError(e.response.data.error.message)
+            case PapieaErrorTypes.Validation:
+                throw new ValidationError(e.response.data.error.message)
+            case PapieaErrorTypes.BadRequest:
+                throw new BadRequestError(e.response.data.error.message)
+            case PapieaErrorTypes.ServerError:
+                throw new PapieaServerError(e.response.data.error.message)
+            default:
+                throw new Error(e)
+        }
+    }
+}
 
 async function create_entity(provider: string, kind: string, version: string, request_spec: Spec, papiea_url: string,meta_extension:any, s2skey: string): Promise<EntitySpec> {
     const payload = {
         spec: request_spec,
         ...(meta_extension) && {metadata: {extension: meta_extension}}
     }
-    const { data: { metadata, spec } } = await axios.post(`${papiea_url}/services/${provider}/${version}/${kind}`, payload, {headers:{"Authorization": `Bearer ${s2skey}`}});
+    const { data: { metadata, spec } } = await make_request(axios.post, `${papiea_url}/services/${provider}/${version}/${kind}`, payload, {headers:{"Authorization": `Bearer ${s2skey}`}});
     return { metadata, spec };
 }
 
 async function create_entity_with_meta(provider: string, kind: string, version: string, meta: Partial<Metadata>, request_spec: Spec, papiea_url: string, s2skey: string): Promise<EntitySpec> {
-    const { data: { metadata, spec } } = await axios.post(`${papiea_url}/services/${provider}/${version}/${kind}`, {
+    const { data: { metadata, spec } } = await make_request(axios.post, `${papiea_url}/services/${provider}/${version}/${kind}`, {
         spec: request_spec,
         metadata: meta
     }, {headers:{"Authorization": `Bearer ${s2skey}`}});
@@ -19,7 +56,7 @@ async function create_entity_with_meta(provider: string, kind: string, version: 
 }
 
 async function update_entity(provider: string, kind: string, version: string, request_spec: Spec, request_metadata: Metadata, papiea_url: string, s2skey: string): Promise<EntitySpec> {
-    const { data: { metadata, spec } } = await axios.put(`${papiea_url}/services/${provider}/${version}/${kind}/${request_metadata.uuid}`, {
+    const { data: { metadata, spec } } = await make_request(axios.put, `${papiea_url}/services/${provider}/${version}/${kind}/${request_metadata.uuid}`, {
         spec: request_spec,
         metadata: {
             spec_version: request_metadata.spec_version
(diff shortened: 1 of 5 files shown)


papiea-client/src/entity_client.ts
```diff
@@ -1,25 +1,62 @@
...
+    ConflictingEntityError,
+    EntityNotFoundError,
+    PermissionDeniedError,
+    ProcedureInvocationError,
+    PapieaServerError,
+    UnauthorizedError,
+    ValidationError
+} from "./errors/errors";
+
+function make_request<T = any>(f: (url: string, data?: any, config?: AxiosRequestConfig) => AxiosPromise<T>, url: string, data?: any, config?: AxiosRequestConfig): AxiosPromise<T> {
```
joshua-berry-ntnx, line 14: style nit: I find long lines unreadable even in editors which do a good job of wrapping them, because the code structure isn't as obvious. Can we please wrap to something sensible (e.g. 80 columns, which is a widely-followed convention dating back to punch cards and typewriters)?

papiea-client/src/errors/errors.ts
```diff
@@ -0,0 +1,31 @@
...
+
+}
+
+export class BadRequestError extends Error {
+
+}
+
+export class PapieaServerError extends Error {
+
+}
```
joshua-berry-ntnx, line 31: Can we add comments explaining these different errors? When/why are they thrown, etc.?

papiea-core/src/core.ts
```diff
@@ -254,13 +254,25 @@ export interface SessionKey {
...
+    Unauthorized,
+    PermissionDenied,
+    ConflictingEntity,
+    ServerError
+}
+
 
 // Modeled after https://developers.google.com/drive/api/v3/handle-errors
 export interface PapieaError {
     error: {
```
joshua-berry-ntnx, line 271: Why are all the fields here wrapped in a single `error` field? This seems redundant.

papiea-client/src/entity_client.ts
Emulebest: I find 80 too small (which isn't a surprise since we have wider screens than those in the typewriters times), a lot of js conventions suggest 120 line, I'd rather go with that

papiea-client/src/errors/errors.ts
Emulebest: Of course

papiea-core/src/core.ts
Emulebest: It was modeled after Google API, IMHO this comes more to a personal preference, I like having an explicit `e.response.data.error`. Let's give @shlomi-nx a final word here

papiea-client/src/entity_client.ts
```diff
@@ -1,25 +1,62 @@
...
+    try {
+        return f(url, data, config)
+    } catch (e) {
+        switch (e?.response?.data?.error.type) {
+            case PapieaErrorTypes.ConflictingEntity:
+                throw new ConflictingEntityError(e.response.data.error.message)
```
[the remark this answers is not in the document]
Emulebest: Agreed

papiea-client/src/entity_client.ts
Emulebest: @joshua-berry-ntnx The problem with papiea is that for some reason we have mixed code styles in different parts of the project. Probably it is because TS/JS doesn't have a unified code style like Python does (PEP8). Maybe I should create a ticket and mark it as `good first issue` to resolve the code style issues. This would be a good starter to newcomers to papiea in Nutanix because it allows to get familiar with the codebase. @shlomi-nx what do you think?

joshua-berry-ntnx: I still find 120 to be too wide. I use my wider screen to have more source files open side-by-side, which is especially helpful when trying to learn a new codebase, understand code flow for debugging, etc. Splitting files horizontally (top/bottom) severely limits the amount of code I can see in a particular file, and wastes more space on the right side of the monitor.

For example, on my current setup at home, I can fit three open files open side-by-side with a bit of room left over for the file tree. Much larger than 80 and I wouldn't be able to do that. (I tested just now, and even 100 doesn't fit, although just barely—this is with 11pt font, btw. :/)

It's a similar situation on my laptop display—I can fit two files side-by-side at 80 columns, but much larger than that and I can only fit one file without resorting to wrapping.

Given that it's a rather arbitrary decision, I've stuck with 80 because it's something of a standard (most editors seem to come preconfigured with 80 as a limit/guideline in some form, and there is long historical precedent), and because it also serves as a nice forcing function to limit expression complexity/verbosity per line (which IME also causes readability issues). I also personally find it easier to read than longer line lengths (where it's harder to scan back to the beginning of the next line).

That said, I'd be willing to explore something like 100 columns if 80 is intolerable.

shlomiv: I can't tell you enough how I hate the 80 cols limit. But then again, if it was up to me, all functions should be one-liners ;)

Unfortunately for us, 80 seems to be the standard. I guess my advise is to keep names short, but not too short.

papiea-core/src/core.ts
shlomi-nx: I think it makes sense to have them in an error field. We can discuss this on the call.

papiea-client/src/entity_client.ts
```diff
@@ -1,119 +1,167 @@
...
 }
 
-export function kind_client(papiea_url: string, provider: string, kind: string, version: string, s2skey?: string, meta_extension?: (s2skey: string)=>any): EntityCRUD {
+export function kind_client(papiea_url: string, provider: string, kind: string, version: string, s2skey?: string, meta_extension?: (s2skey: string) => any): EntityCRUD {
     const the_s2skey = s2skey ?? 'anonymous'
     const crudder: EntityCRUD = {
         get: (entity_reference: Entity_Reference) => get_entity(provider, kind, version, entity_reference, papiea_url, the_s2skey),
-        create: (spec: Spec) => create_entity(provider, kind, version, spec, papiea_url, meta_extension ? meta_extension(the_s2skey):undefined, the_s2skey),
-        create_with_meta: (meta: Partial<Metadata>, spec: Spec) => create_entity_with_meta(provider,kind, version, meta, spec, papiea_url, the_s2skey),
+        create: (spec: Spec) => create_entity(provider, kind, version, spec, papiea_url, meta_extension ? meta_extension(the_s2skey) : undefined, the_s2skey),
```
shlomi-nx, line 140: also spaces between `:` and `=>`. I don't see the purpose. Lets get rid of these.

papiea-core/src/core.ts
```diff
@@ -254,13 +254,25 @@ export interface SessionKey {
     idpToken: any
 }
 
+export enum PapieaErrorTypes {
```
shlomi-nx, line 257: This should be singular `PapieaErrorType`

shlomi-nx: Perhaps even just `PapieaError`

papiea-client/src/entity_client.ts
joshua-berry-ntnx: Conventional JS style writes these like `foo: Bar` and `(arg) => ret`, so you will confuse native JS speakers.

shlomi-nx: Please allow me to 🤦 ^ 3 (note the spaces). Fine. Will the delicate natives be able to not have a space wrapping every curly?

joshua-berry-ntnx: Yes I think that's fine =P

papiea-core/src/core.ts
Emulebest: There is already a type `PapieaError` right below it :) It is an interface for JSON response error

PR #379, opened 2020-05-05
Fixed update status endpoint to include prefix & version

### LINKED ISSUE #282: Update status endpoint should include prefix & version

Emulebest:
resolves #282

### DIFF
--- a/papiea-backend-utils/src/logging.ts
+++ b/papiea-backend-utils/src/logging.ts
@@ -46,7 +46,6 @@ export class LoggerFactory {
     readonly options: LoggerOptions
 
     private static readonly INSPECT_OPTIONS = {
-        colors: true, depth: 3, maxStringLength: 120,
         sorted: true, getters: true
     }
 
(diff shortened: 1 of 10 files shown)


papiea-backend-utils/src/logging.ts
```diff
@@ -46,7 +46,6 @@ export class LoggerFactory {
     readonly options: LoggerOptions
 
     private static readonly INSPECT_OPTIONS = {
-        colors: true, depth: 3, maxStringLength: 120,
```
shlomi-nx, line 49: Already seen this in another PR.. Why are we removing colors?

Emulebest: It's the 2nd PR where I keep forgetting to address this change. The coloring seems to generate exceptions on my machine (I suspect it's because some optional dep being ommited), which isn't a big problem by itself. The problem is that coloring exceptions prevent all other logs from working. I think I'll file a ticket to introduce safer coloring and fix bugs on some machines

PR #386, opened 2020-05-12
Fix weird messages during validation errors

### LINKED ISSUE #382: Swagger schema validation errors sometimes don't provide useful information
In BMaaS, I have defined an entity procedure which takes no arguments. Paraphrased:

```ts
    const args = {type: 'object', properties: {}}
    machine_kind.entity_procedure('deprovision', {}, Halt_Intentful, {input: args}, {output: {}}, fn)
```

I call the entity procedure like so:

```ts
await machine.invoke_procedure('deprovision', machine, {})
```

But the entity procedure does not even get called (I have log messages at the beginning of the procedure which don't fire). Instead I get the following from papiea-engine:

```
bmaas_papiea | {"0":"Error: [{\"message\":\"Unable to validate an empty value for property: rootModel\"}]\n    at Function.fromError (/code/papiea-engine/src/errors/procedure_invocation_error.ts:26:20)\n    at Entity_API_Impl.call_procedure (/code/papiea-engine/src/entity/entity_api_impl.ts:166:44)\n    at process._tickCallback (internal/process/next_tick.js:68:7)","level":"error","message":{"error":{"code":400,"errors":[{"message":"Unable to validate an empty value for property: rootModel"}],"message":"Procedure invocation failed."}},"timestamp":"2020-05-07T19:03:18.375Z"}
```

Presumably (I haven't checked yet and am just guessing) the problem is my output schema is an empty object. But I have no idea what `rootModel` is and the error doesn't point to the specific part of my code which is causing the problem.

I've seen this same `rootModel` error in a few other contexts—for example, when passing ill-formed metadata to the engine REST API (e.g. when trying to PUT an entity to set an intent). We should try to make this error more specific and reference the specific thing that the user actually provided which is causing the problem.

### LINKED ISSUE #383: Validation always fails when a procedure takes an empty object as input
[Note: This uses a similar but slightly different example from #382.]

In BMaaS, I have defined an entity procedure which takes no arguments. Paraphrased:

```ts
    const args = {type: 'object', properties: {}}
    machine_kind.entity_procedure('deprovision', {}, Halt_Intentful,
        {input: args}, {output: undefined}, fn)
```

The reason I want to do this is deprovision() takes no arguments now, but I want to allow for arguments to be added later if necessary.

I call the entity procedure like so:

```ts
await machine_client.invoke_procedure('deprovision', some_machine, {})
```

But the entity procedure does not even get invoked and `400 Bad Request` is returned to the client, with papiea-engine complaining that: `Unable to validate an empty value for property: rootModel`

I've experimented a bit, and the problem seems to be the Swagger definition. If I pass `undefined` in place of `args` above, everything works (and, oddly, I can still call the procedure with an empty object `{}` and validation passes). But if I define an empty object in the Swagger schema like in the example above, I get the error regardless of whether I pass `{}` or `undefined` when calling the procedure.

I would expect that defining an `undefined` Swagger schema would only accept `undefined` as a parameter (and not `{}`), and conversely, defining an object with no properties would accept `{}` (but not `undefined`).

I can work around this for now by simply passing `undefined` for the Schema, but this seems like an oddity we may want to look into further.

Emulebest:
resolves #383 
resolves #382


papiea-engine/src/validator/index.ts
```diff
@@ -15,7 +23,17 @@ export class ValidatorImpl {
...
         const validatorDenyExtraProps = !allowExtraProps
         const allowBlankTarget = false
-        if (model) {
+        if (modelIsEmpty(model)) {
+            if (isEmpty(data)) {
+                return {valid: true}
+            } else {
+                throw new ValidationError([{
+                    name: "Error",
+                    message: "Function was expecting empty object"
```
shlomi-nx, line 32: An Empty object as an input argument? Is there a way to get the swagger name of that model to show the user? I would also print the function name (procedure name).

More information makes users happier.

PR #388, opened 2020-05-13
NPM giving us random problem

shlomi-nx:
NPM is giving us random problems, package-locks causes random things to fail, random packages are installed as root (even though the npm command was not ran as root) etc.


Makefile
```diff
@@ -60,8 +60,14 @@ clean-node-modules:
...
 
+clean: clean-build
 
 clean-build:
 	npm run clean-all; \
 	cd ./papiea-engine/__benchmarks__; \
 	rm -rf build
+
+clean-package-lock:
+	for p in core client backend-utils engine sdk; do rm -rf papiea-$$p/package-lock.json; done; \
```
shlomi-nx, line 71: @joshua-berry-ntnx Should we extract this list of packages to some constant? This is being used in a few targets

Makefile
```diff
@@ -34,7 +34,7 @@ node_modules:
 
 stop-papiea:
 		cd ./papiea-engine; \
-    	if [ -z `docker-compose ps -q papiea-engine` ] || [ -z `docker ps -q --no-trunc | grep $$(docker-compose ps -q papiea-engine)` ]; then \
+        if [ -z `docker-compose ps -q papiea-engine` ] || [ -z `docker ps -q --no-trunc | grep $$(docker-compose ps -q papiea-engine)` ]; then \
```
joshua-berry-ntnx, line 37: whitespace-only change?

Makefile
```diff
@@ -60,8 +60,14 @@ clean-node-modules:
 	cd ./papiea-engine/__benchmarks__; \
 	rm -rf node_modules
 
+clean: clean-build
```
joshua-berry-ntnx, line 63: I think it's fine to just rename `clean-build` to `clean`.

Makefile
joshua-berry-ntnx: Yes, I think this is a good idea. You could just define the space-separated list of packages as a make variable, or if you want to get fancy, you could define a variable which expands to the full `papiea-core papiea-client ...` list of package names using something like `$(patsubst %,papiea-%,core client ...)`.

Makefile
shlomi-nx: Emacs did not like the previous whitespace. I don't argue with emacs.

shlomi-nx: You will love my next commit, a bunch of whitespace-only changes!

joshua-berry-ntnx: Yes, I know from experience that emacs is very good at catching whitespace problems in makefiles. I will defer to emacs here. :)

Emulebest:
closed because of yarn introduction in #391

PR #391, opened 2020-05-15
[WIP] Yarn instead of Npm

Emulebest:
connected to #388 

There are a few things to notice:
Yarn is a lot more straightforward than npm, thus it has different way of solving "papiea-core: file://" dependencies, more on that below.
Yarn simply copies the "papiea-core: file://" dep folder to where it is needed in node_modules, so the build order should be a little different, an example:
Papiea-engine depends on papiea-core and papiea-backend-utils, the build order for that should be:
1) yarn install on papiea-core
2) yarn run build on papiea-core (build directory appears)
3) yarn install on papiea-backend-utils (copies the build directory from papiea-core which is need for definitons)
4) yarn run build on papiea-backend-utils (build directory appears)
5) yarn install on papiea-engine (copies both build dirs from papiea-core and papiea-backend-utils)
6) yarn run build on papiea-engine


shlomi-nx:
We should capture all these in the makefiles

package.json
```diff
@@ -5,12 +5,13 @@
...
-    "build-all": "for p in core client backend-utils engine sdk; do npm --prefix papiea-$p run build || exit 1; done",
-    "patch-all": "for p in core client backend-utils engine sdk; do npm --prefix papiea-$p version patch || exit 1; done",
-    "install-all": "for p in core client backend-utils engine sdk; do npm --prefix papiea-$p install || exit 1; done",
-    "build-all!": "npm run clean-all && npm run build-all",
-    "publish-all": "npm run build-all! && node publish-papiea.js"
+    "upgrade-all": "for p in core client backend-utils engine sdk; do yarn --cwd papiea-$p upgrade || exit 1; done",
+    "clean-all": "for p in core client backend-utils engine sdk; do yarn --cwd papiea-$p run clean || exit 1; done",
+    "build-all": "for p in core client backend-utils engine sdk; do yarn --cwd papiea-$p run build || exit 1; done",
+    "patch-all": "for p in core client backend-utils engine sdk; do yarn --cwd papiea-$p version patch || exit 1; done",
+    "install-all": "for p in core client backend-utils engine sdk; do yarn --cwd papiea-$p || exit 1; done",
```
shlomi-nx, line 12: Perhaps these should be rewritten in terms of the Makefile (might need to add targets)

README.md
```diff
@@ -49,7 +47,7 @@ See [Attaching to node js using Intellij](https://www.jetbrains.com/help/pycharm
 
 1. From project root `cd papiea-engine/__benchmarks__`
 
-2. Install dependencies by `npm install`
+2. Install dependencies by `yarn` (Note you need to have yarn installed)
```
shlomi-nx, line 50: Perhaps a link to yarn installation tutorial would be useful here

papiea-engine/package.json
```diff
@@ -54,7 +54,8 @@
     "swagger-ui-express": "^4.0.7",
     "uuid": "^3.3.2",
     "uuid-validate": "0.0.3",
-    "wait-port": "^0.2.2"
+    "wait-port": "^0.2.2",
+    "ts-node": "^7.0.1"
```
shlomi-nx, line 58: Why ts-node moved from `devDependencies`?

Also, why do we need `wait-port`? @joshua-berry-ntnx simply used `GET /` instead and I think I like it better

papiea-engine/start.sh
```diff
@@ -7,7 +7,6 @@ fi
 
 DB_HOST=${MONGO_HOST:-'mongo'}
 DB_PORT=${MONGO_PORT:-'27017'}
-npm install
 npm run build-clj
```
shlomi-nx, line 10: is npm used here only to build clojure part?

papiea-engine/package.json
Emulebest: Apparently it is used to do typescript to javascript and run the js, yarn think it is a dependency which might be true since we use it in runtime to start

README.md
Emulebest: Sure

papiea-engine/start.sh
Emulebest: Yes, npm is still used in couple places. BUT I limited it to be used only in scenarios where we don't touch deps. Like starting tests in the container (not to install yarn) and build clojure

package.json
Emulebest: I think these are good to be here. Makefiles already do the appropriate build order to our project. I'd leave those in package.json as is because they are more or less generic and general purpose

Makefile
```diff
@@ -21,15 +23,50 @@ run-papiea: papiea-build
...
+		docker-compose logs --tail=5 papiea-engine | grep 'app listening on port' && echo Success && exit 0; \
+		sleep 2; \
+		docker-compose logs --tail=5; \
+	  done \
+	fi
 
-papiea-build: node_modules
-	npm run build-all
 
+build_all: build_bench
```
joshua-berry-ntnx, line 39: Should this also depend on `build_main` and `build_deps`?

Makefile
```diff
@@ -21,15 +23,50 @@ run-papiea: papiea-build
...
+
+
+node_modules_main: build_deps
+	yarn --cwd papiea-engine; \
+	yarn --cwd papiea-sdk
+
+
+build_main: node_modules_main
+	yarn --cwd papiea-engine run build; \
+	yarn --cwd papiea-sdk run build
```
joshua-berry-ntnx, line 61: Let's pick a consistent order for sorting make targets, otherwise the Makefile will become hard to read; I suggest putting dependent targets above all their dependencies (so build_main would appear before node_modules_main).

README.md
```diff
@@ -125,7 +123,7 @@ For the third one, and this one is really strange, some of the folders in `npm_m
 1. `sudo make clean-all`
 3. `make` - should now pass
 
-We are looking into moving away from `npm`, as these troubles are happening randomly and without an appearnt cause. 
+We are looking into moving away from `npm`, possibly to `yarn`, as these troubles are happening randomly and without an appearnt cause.
```
joshua-berry-ntnx, line 126: typo: "apparent"

Makefile
```diff
@@ -21,15 +23,50 @@ run-papiea: papiea-build
...
+
+
+node_modules_bench: build_main
+	yarn --cwd papiea-engine/__benchmarks__; \
```
joshua-berry-ntnx, line 65: For these `node_modules_foo` targets, we should probably name the actual `node_modules` directories they're pointing to (e.g. `papiea-engine/__benchmarks__/node_modules`), and dependencies on `package.json` (or whatever) so make knows the relationships and can avoid running yarn at all if it's not necessary.

README.md
joshua-berry-ntnx: `brew install yarn` works on OSX, for the record.

package.json
shlomi-nx: I am concerned with the lists `[core, client, ...]` needing maintenance in Makefile as well as all these targets individually. This can get out of sync easily.

papiea-engine/package.json
shlomi-nx: Ok, what about wait-port?

Emulebest: Wait port could be removed

shlomi-nx: Great, please remove it

README.md
shlomi-nx: Lets add a link to https://classic.yarnpkg.com/en/docs/install/#debian-stable here

PR #399, opened 2020-05-29
Сustomizable uuid

### LINKED ISSUE #368: [Enhancement] - Make metadata.uuid format customizable per kind
# Background
This proposal is concerned with the way we can identify entities. Papiea currently uses opaque uuid4 exclusively, which although convenient in terms of the engine, may not be that convenient in terms of usability.

An Identifier can have the following attributes:

1. Immutability - an Identifier should remain consistent so it is possible to refer to the same identity over time
2. Uniqueness - an Identifier must be unique in a relevant domain, so that it can identify a single identity
3. Meaningfulness - an Identifier should strive to be meaningful to users consuming the system. This improves the ability to reason about a system.

While the first two items are mandatory, the third is optional and is convenient to have. However, it can clash with the first two:
1. Meaningfulness clash with Immutability - something meaningful now can be meaningless later. If selecting a meaning has to happen at the time of creation, the meaning may be too abstract, or outright incorrect (e.g. `dev-vm` which may turn into a `prod-vm`). If we cannot refresh the meaning, than it is nothing but confusing.
2. Meaningfulness clash with Uniqueness - how many `dev-vm`'s can a system have? Just one?

Since Papiea is multi-tenant in a few dimensions, i.e. multiple providers, multiple versions, multiple users, dynamic roles on users etc., Uniqueness has a varying domain. Do we need Uniqueness only within a kind, but across users and providers? Do we need Uniqueness per role? It is rather hard to answer.

In order to address the following issues, we identified two general cases of kinds:
1. Kinds that are instantiable by users, e.g. vm's
2. Kinds that are a representation of a system, and are only instantiated by the provider, e.g. instance types.

The first kind type are usually good for using uuids, as a meaningful name may either be hard to come up with, or may not stand the test of time. Also, determining uniqueness boundary may be difficult.

The second kind type can have meaningful identifier's as they contain very specific meaning that is not bound to change. E.g. instance-types, an `i3.metal` is a meaningful identifier, that is not going to change. 

# Proposal
For this purpose, we propose the following solution:
1. Provider may define Kinds as usual, with the default uuid format as standard uuid4
2. When a Provider defines a Kind, it may optionally provide a regular expression that accepts valid identifiers for instances of this particular kind
3. When an entity of such kind is created, a uuid MUST be provided (since Papiea would not know how to generate a valid one, let alone a meaningful one. The provided uuid is then checked for two things:
   1. The uuid conforms to the provided regex
   2. The uuid is unique for the entire provider boundary
4. When the pre-checks above are passed the instance is created, where its `metadata.uuid` contains the given, meaningful, immutable and unique identifier.
5. Since the regex is known at registration time, it could be automatically documented in the exported swagger specs.

# Why
## Usability
For the second type of kinds (defined above), it may be more convenient for users to refer to an instance by such meaningful uuid, e.g.:
```
const new_vm = vm.create({ ... ,
                          instance_type_ref: {kind:'instance_type',
                                              uuid:'i3.metal'},
                          ...})
```
Instead of going through the more traditional:
```
const i3metal = instance_types.filter({'spec.name':'i3.metal'})[0]
const new_vm = vm.create({ ... ,
                          instance_type_ref: ref(i3metal)},
                          ...})
```
## Generability
Since Papiea is a general purpose container for services, it may be desired by different services to establish their own identity semantics.

Emulebest:
resolves #368


shlomi-nx:
@Ntnx-Igor-Korotach please resolve the newly introduced conflicts and merge

Emulebest:
Swagger now features pattern for uuid validation
image

PR #410, opened 2020-06-05
Reregister provider feature

### LINKED ISSUE #405: Papiea does not permit registering an already registered provider/version
Papiea used to allow us to override a provider/version, e.g. `griffon/0.1.0`. Now, if we try to override it we are getting:

```
api - check_response: Got exception while makeing request. Status: 500, Reason: Internal Server Error, Details:{
    "error":{
        "code":500,
        "errors":[
            {
                "message":"E11000 duplicate key error collection: papiea.intentful_kinds index: provider_prefix_1_provider_version_1_kind_name_1 dup key: { provider_prefix: \"griffon\", provider_version: \"0.1.0\", kind_name: \"node\" }"
            }
        ],
        "message":"E11000 duplicate key error collection: papiea.intentful_kinds index: provider_prefix_1_provider_version_1_kind_name_1 dup key: { provider_prefix: \"griffon\", provider_version: \"0.1.0\", kind_name: \"node\" }",
        "type":7
    }
}
```

We should allow overriding an existing version. If we would like to disallow it, it will need to be some configuration on papiea (say, for production deployements). 

For now lets simply allow it.

Emulebest:
resolves #405

### DIFF
--- a/papiea-engine/__tests__/provider_tests/provider_api.test.ts
+++ b/papiea-engine/__tests__/provider_tests/provider_api.test.ts
@@ -41,6 +41,15 @@ describe("Provider API tests", () => {
         providerApi.post('/', provider).then(() => done()).catch(done.fail);
     });
 
+    test("ReRegister provider", done => {
+        const prefix = "test_provider_reregister"
+        const original_provider: Provider = { prefix: prefix, version: providerVersion, kinds: [], procedures: {}, extension_structure: {}, allowExtraProps: true };
+        providerApi.post('/', original_provider).then(() => done()).catch(done.fail);
+        const overriden_provider: Provider = { prefix: prefix, version: providerVersion, kinds: [], procedures: {}, extension_structure: {}, allowExtraProps: false };
+        providerApi.post('/', overriden_provider).then(() => done()).catch(done.fail);
+        providerApi.delete(`/${prefix}/${providerVersion}`).then(() => done()).catch(done.fail);
+    });
+
     test("Register malformed provider", done => {
         providerApi.post('/', {}).then(() => done.fail()).catch(() => done());
     });
--- a/papiea-engine/src/databases/provider_db_mongo.ts
+++ b/papiea-engine/src/databases/provider_db_mongo.ts
@@ -65,7 +65,18 @@ export class Provider_DB_Mongo implements Provider_DB {
         if (intentful_kinds.length === 0) {
             return
         }
-        await this.subCollection.insertMany(intentful_kinds)
+        try {
+            await this.subCollection.insertMany(intentful_kinds)
+        } catch (err) {
+            // Check a DuplicateMongo Error
+            // If the error is a DuplicateMongo Error
+            // then we are reregestering provider and its fine
+            if (err.code === 11000) {
+                return
+            } else {
+                throw err
+            }
+        }
     }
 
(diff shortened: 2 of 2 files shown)


papiea-engine/__tests__/provider_tests/provider_api.test.ts
```diff
@@ -41,6 +41,15 @@ describe("Provider API tests", () => {
         providerApi.post('/', provider).then(() => done()).catch(done.fail);
     });
 
+    test("ReRegister provider", done => {
+        const prefix = "test_provider_reregister"
+        const original_provider: Provider = { prefix: prefix, version: providerVersion, kinds: [], procedures: {}, extension_structure: {}, allowExtraProps: true };
+        providerApi.post('/', original_provider).then(() => done()).catch(done.fail);
+        const overriden_provider: Provider = { prefix: prefix, version: providerVersion, kinds: [], procedures: {}, extension_structure: {}, allowExtraProps: false };
+        providerApi.post('/', overriden_provider).then(() => done()).catch(done.fail);
+        providerApi.delete(`/${prefix}/${providerVersion}`).then(() => done()).catch(done.fail);
```
shlomi-nx, line 50: The test only checks that no exception was thrown, not that the provider actually got updated

papiea-engine/src/databases/provider_db_mongo.ts
```diff
@@ -65,7 +65,18 @@ export class Provider_DB_Mongo implements Provider_DB {
...
         }
-        await this.subCollection.insertMany(intentful_kinds)
+        try {
+            await this.subCollection.insertMany(intentful_kinds)
+        } catch (err) {
+            // Check a DuplicateMongo Error
+            // If the error is a DuplicateMongo Error
+            // then we are reregestering provider and its fine
+            if (err.code === 11000) {
+                return
```
shlomi-nx, line 75: Is this correct? Is the entity getting updated and throws the exception after a successful update? This looks only like ignoring the exception.

Emulebest:
```
            await this.subCollection.createIndex({
                "provider_prefix": 1,
                "provider_version": 1,
                "kind_name": 1
            }, { unique: true })
```
This the piece of code that throws since we are making a unique collection of intenful kinds and `save_provider` tries to do that once again. Considering that kinds didn't change it is safe that we keep the old index, if they had changed then it is not going to raise an Exception since a new set of unique pair is going to be generated

papiea-engine/__tests__/provider_tests/provider_api.test.ts
Emulebest: I see, though this should work since we had a logic to reregister provider in the past and its only with the recent addition of subcollection below that it started throwing. Nevertheless I'll add the test but this is ought to be working even now with this PR

papiea-engine/src/databases/provider_db_mongo.ts
shlomi-nx: hmm.. why are we recreating the index? Isn't mongo updating the index when content changes?

Emulebest: I might be a bit wrong on this one but Mongo doesn't really modify the index at all, you need to reindex for that (drop -> recreate)

PR #414, opened 2020-06-09
Replace status done & tested, PATCH update_status has some issues

### LINKED ISSUE #411: Nested objects contained in arrays are serialized improperly
For example:

```js
ctx.update_status(some_entity, {
    nonsense: [{
        a: {b: 'c', d: 'e'},
        f: 'g'
    }]
})
```

Will actually return this when retrieving the status using a GET request

```js
status: {
    "nonsense": [{
        "a.b": "c",
        "a.d":"e",
        "f":"g"
    }]
}
```

I'm not sure if this applies to spec as well as status, and I haven't been able to check what's actually stored in Mongo (not sure how), but there definitely seems to be some issue in the serialization/deserialization path somewhere.

Emulebest:
resolves #411

### DIFF
--- a/papiea-python-sdk/papiea/python_sdk_context.py
+++ b/papiea-python-sdk/papiea/python_sdk_context.py
@@ -88,6 +88,15 @@ class ProceduralCtx(object):
             {"entity_ref": entity_reference, "status": status},
         )
 
+    async def replace_status(
+        self, entity_reference: EntityReference, status: Status
+    ):
+        url = f"{self.provider.get_prefix()}/{self.provider.get_version()}"
+        await self.provider_api.post(
+            f"{url}/update_status",
+            {"entity_ref": entity_reference, "status": status},
+        )
+
     def update_progress(self, message: str, done_percent: int) -> bool:
         raise Exception("Unimplemented")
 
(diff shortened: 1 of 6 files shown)


papiea-python-sdk/papiea/python_sdk_context.py
```diff
@@ -88,6 +88,15 @@ async def update_status(
...
 
+    async def replace_status(
+        self, entity_reference: EntityReference, status: Status
+    ):
+        url = f"{self.provider.get_prefix()}/{self.provider.get_version()}"
+        await self.provider_api.post(
+            f"{url}/update_status",
+            {"entity_ref": entity_reference, "status": status},
+        )
+
```
shlomi-nx, line 99: Isn't this too low level? This may cause users to introduce bugs.

Ntnx-Igor-Korotach: I recommend merging this as this is an option for providers to avoid bugs as now we have none built in SDK.
Regarding ergonomics and our discussion, I'll tackle those in https://github.com/nutanix/papiea/issues/415

PR #419, opened 2020-06-11
Specific Python-SDK papiea errors

### LINKED ISSUE #402: Procedure returns a numeric error code
Same instructions as #401, this time authenticated and created an entity with proper extension. Trying to invoke a procedure gives:

```
{
  "error": {
    "code": 500,
    "errors": [
      {}
    ],
    "message": "Procedure invocation failed.",
    "type": 2
  }
}
```

No details, and `type:2` are really not helpful to users.

### LINKED ISSUE #404: python-sdk - update_status should return the failure reason
Trying to set status with wrong field shows `validation error` on engine's side, but not on the sdk 
(provider) side:

https://github.com/nutanix/papiea/blob/85362f7a26b3cb878c9c76fb5be04a33290cdd30/papiea-python-sdk/papiea/python_sdk_context.py#L80-L82

This should not return a bool, rather should throw an error with the appropriate message

### LINKED ISSUE #409: Add more specific exceptions to Papiea SDK

Emulebest:
resolves #409 
resolves #404 
resolves #402

### DIFF
--- a/papiea-python-sdk/papiea/api.py
+++ b/papiea-python-sdk/papiea/api.py
@@ -11,9 +11,10 @@ from papiea.utils import json_loads_attrs
 
 
 class ApiInstance(object):
-    def __init__(self, base_url: str, timeout: int = 5000 * 60, headers: dict = {}, *, logger: logging.Logger):
+    def __init__(self, base_url: str, timeout: int = 5000, headers: dict = {}, *, logger: logging.Logger):
         self.base_url = base_url
         self.headers = headers
+        # Shlomi: This does not work! 
         self.session = ClientSession(timeout=ClientTimeout(total=timeout))
         self.logger = logger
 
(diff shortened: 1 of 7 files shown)


shlomi-nx:
@Emulebest Please check out my last two commits. Stashing the session was the culprit. The session-renew part makes this run properly. We should probably add this to the rest of the api calls in api.py

papiea-python-sdk/papiea/api.py
```diff
@@ -58,68 +34,117 @@ async def post(self, prefix: str, data: dict, headers: dict = {}) -> Any:
...
+                return None
+            return json_loads_attrs(res)
+        except PapieaBaseException as papiea_exception:
+            raise papiea_exception
+        except ApiException as api_exception:
+            raise api_exception
+        except Exception as e:
+            self.logger.debug("RENEWING SESSION")
+            await self.renew_session()
+            await self.patch(prefix, data, headers)
```
shlomi-nx, line 99: This is not a good solution, as if there is some unexpected exception, this will recurse until it will reach stackoverflow. The reason I had the closure in the original fix was that it will only get retried once. If the problem is with a stale connection, a single renew must solve it. It if did not solve it - the error should propagate to the caller.

Emulebest: Good catch, didn't think of it that way. I'll abstract the call to `post, patch, etc..` to a generic helper function

PR #420, opened 2020-06-11
Fix update_status to have vector-replace semantics

### LINKED ISSUE #415: PATCH update_status doesn't work well with arrays

shlomi-nx:
Add dotnotation which skips vectors

Fixes #415

### DIFF
--- a/papiea-backend-utils/src/dotnotation.ts
+++ b/papiea-backend-utils/src/dotnotation.ts
@@ -1,0 +1,75 @@
+// Shlomi: Moved encode here from mongo-dot-notation-tool and fixed it to not venture into vectors
+// Need to add tests to this
+function encode(value: any, keyChain: string[], result: any) {
+    const isObject = (value: any)=> value && value.toString() === '[object Object]'
+    const isArray = (value: any)=>  Array.isArray(value);
+    let _key:any;
+
+    if (!keyChain) {
+        keyChain = [];
+    }
+
+    if (isArray(value)) {
+        if (!result) {
+            result = [];
+        }
+
+        for (var i = 0; i < value.length; i++) {
+            result[i] = value[i];
+        }
+    } else if (isObject(value)) {
+        if (!result) {
+            result = {};
+        }
+
+        Object.keys(value).forEach(function (key) {
+            let _keyChain = ([] as string[]).concat(keyChain);
+            _keyChain.push(key);
+
+            if (key.charAt(0) === '$') {
+                if (isArray(value[key])) {
+                    result[key] = value[key];
+                } else if (isObject(value[key])) {
+                    if (keyChain.length) {
+                        _key = keyChain.join('.');
+                        if (!result[_key]) {
+                            result[_key] = {};
+                        }
+                        encode(value[key], [key], result[_key]);
+                    } else {
+                        encode(value[key], [key], result);
+                    }
+                } else {
+                    _key = keyChain.join('.');
+                    let _o: any = {};
+                    _o[key] = value[key] as any;
+                    if (result[_key]) {
+                        Object.assign(result[_key], _o);
+                    } else {
+                        result[_key] = _o;
+                    }
+                }
+            } else {
+                if (isArray(value[key])) {
+                    _key = _keyChain.join('.');
+                    result[_key] = [];
+                    encode(value[key], [], result[_key]);
+                } else if (isObject(value[key])) {
+                    encode(value[key], _keyChain, result);
+                } else {
+                    result[_keyChain.join('.')] = value[key];
+                }
+            }
+        });
+    } else {
+        result = value;
+    }
+
+    return result;
+}
+
+const dotnotation = (x:any)=>encode(x, [], undefined)
+
+export {
+    dotnotation
+}
--- a/papiea-backend-utils/src/dotnotation.ts
+++ b/papiea-backend-utils/src/dotnotation.ts
@@ -1,7 +1,7 @@
 // Shlomi: Moved encode here from mongo-dot-notation-tool and fixed it to not venture into vectors
 // Need to add tests to this
 function encode(value: any, keyChain: string[], result: any) {
-    const isObject = (value: any)=> (value instanceof Object && !(value instanceof Array))
+    const isObject = (value: any)=> value && value.toString() === '[object Object]'
     const isArray = (value: any)=>  Array.isArray(value);
     let _key:any;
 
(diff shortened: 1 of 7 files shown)


papiea-backend-utils/src/dotnotation.ts
```diff
@@ -0,0 +1,75 @@
+// Shlomi: Moved encode here from mongo-dot-notation-tool and fixed it to not venture into vectors
+// Need to add tests to this
+function encode(value: any, keyChain: string[], result: any) {
+    const isObject = (value: any)=> value && value.toString() === '[object Object]'
```
joshua-berry-ntnx, line 4: Why is `typeof value === 'object'` not sufficient?

papiea-backend-utils/src/dotnotation.ts
```diff
@@ -0,0 +1,75 @@
...
+    } else if (isObject(value)) {
+        if (!result) {
+            result = {};
+        }
+
+        Object.keys(value).forEach(function (key) {
+            let _keyChain = ([] as string[]).concat(keyChain);
+            _keyChain.push(key);
+
+            if (key.charAt(0) === '$') {
```
joshua-berry-ntnx, line 29: I don't understand this special handling for keys beginning with `$`. Why is this needed and can you add a comment?

papiea-engine/src/databases/status_db_mongo.ts
```diff
@@ -41,12 +40,12 @@ export class Status_DB_Mongo implements Status_DB {
...
 
     async update_status(entity_ref: Entity_Reference, status: Status): Promise<void> {
-        const partial_status_query = encode({"status": status});
+        const partial_status_query = dotnotation({"status": status});
         const result = await this.collection.updateOne({
             "metadata.uuid": entity_ref.uuid,
             "metadata.kind": entity_ref.kind
         }, {
-                $set: partial_status_query
+                $set: partial_status_query
```
joshua-berry-ntnx, line 48: whitespace-only change

papiea-backend-utils/src/dotnotation.ts
shlomi-nx: This piece of code is copied and modified from the original tool. I did not go over the entire algorithm implemented, just wanted to adapt what we were already using. I suppose this pertains to mongo's special `$..` commands.

papiea-backend-utils/src/dotnotation.ts
shlomi-nx: This is copied from the original tool. Although a few stackoverflows has suggested the same thing.

papiea-engine/src/databases/status_db_mongo.ts
shlomi-nx: Thats what a purely green rectangle means, yes. :P

papiea-backend-utils/src/dotnotation.ts
shlomi-nx: If I think about it, it may be because a vector is also considered as an object, and we need to distinguish between the two.

Emulebest: Js Arrays are also typedef Objects, that's why. In general JS typesystem is ... a bit off, so this is good as far as it could go

shlomi-nx:
> LGTM, though I would consider writing/copying tests from mongo-dot-notation tool to check for edge cases

Of course.

papiea-backend-utils/src/dotnotation.ts
Emulebest: That's for $or, $and, etc. mongo queries

papiea-engine/src/databases/status_db_mongo.ts
joshua-berry-ntnx: Think of your laptop's poor overtaxed SSD before wasting bits on trailing spaces. 🤣

papiea-backend-utils/src/dotnotation.ts
joshua-berry-ntnx: So the code always seems to call `isArray` before calling `isObject`... and I suspect `typeof value === 'object' && ! Array.isArray(value)` is probably still more efficient (and honestly, safer) than calling `toString()` on arbitrary objects.

shlomi-nx: Sure, this can be changed. However, as I said, that piece of code was copied from what we already used. I suppose the review I had in mind here is the review of the changes I introduced to the code, not the code itself.

In any event, I would not rely on the apparent ordering of calls by the caller. The second option you proposed is much better.

papiea-engine/src/databases/status_db_mongo.ts
shlomi-nx: don't cry over spilled bits? lol 

I am much more concerned with all the leading spaces on every single line! This is just "shot noise" 😆

papiea-backend-utils/src/dotnotation.ts
shlomi-nx: I have been looking around, and it seems that the toString thing is pretty expected.. The highly used jquery uses the same!

https://github.com/jquery/jquery/blob/d0ce00cdfa680f1f0c38460bc51ea14079ae8b07/src/core.js#L204-L223

```javascript
	isPlainObject: function( obj ) {
		var proto, Ctor;


		// Detect obvious negatives
		// Use toString instead of jQuery.type to catch host objects
		if ( !obj || toString.call( obj ) !== "[object Object]" ) {
			return false;
		}


		proto = getProto( obj );


		// Objects with no prototype (e.g., `Object.create( null )`) are plain
		if ( !proto ) {
			return true;
		}


		// Objects with prototype are plain iff they were constructed by a global Object function
		Ctor = hasOwn.call( proto, "constructor" ) && proto.constructor;
		return typeof Ctor === "function" && fnToString.call( Ctor ) === ObjectFunctionString;
	},
```

I would keep what we have..

shlomi-nx:
The last commit `Saner isObject` is not sufficiently good. Thinking of reverting.

papiea-backend-utils/src/dotnotation.ts
joshua-berry-ntnx: Actually the way jQuery does it is (somewhat) better, because it's calling the global `toString()` rather than relying on method dispatch which can be influenced by the object itself. (But jQuery's version won't work in Node, which according to my testing returns a different string and doesn't seem to actually be useful for anything except returning the constant `"[object Undefined]"`...)

In any case, I'm willing to let it pass for now since it's copy/paste...

PR #425, opened 2020-06-16
fix and test for delay time

### LINKED ISSUE #424: Rediff delay add test

Emulebest:
resolves #424

### DIFF
--- a/papiea-sdk/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
+++ b/papiea-sdk/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
@@ -243,7 +243,7 @@ describe("Intentful Workflow tests", () => {
                     })
                 } else {
                     // 12 seconds delay
-                    return 12
+                    return {"delay_secs": 12}
                 }
             })
             location.on_create(async (ctx, entity) => {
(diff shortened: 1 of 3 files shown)


papiea-sdk/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
```diff
@@ -220,6 +220,72 @@ describe("Intentful Workflow tests", () => {
...
+                        context: "some context",
+                        entity_ref: {
+                            uuid: metadata.uuid,
+                            kind: kind_name
+                        },
+                        status: { x: entity.spec.x }
+                    })
+                } else {
+                    // 12 seconds delay
+                    return 12
```
shlomi-nx, line 246: Could we change the return type to be an object, something like:
```js
return {"delay_secs": 12}
```

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -482,7 +482,7 @@ export class Kind_Builder {
                 IntentfulOutput: {
                     type: 'integer',
                     nullable: true,
-                    description: "Delay for intentful operation"
+                    description: "Delay for the entity to be rediffed in (measured in seconds)"
```
shlomi-nx, line 485: `Amount of seconds to wait before this entity will be checked again by the intent engine`

PR #426, opened 2020-06-16
[WIP] Introducing tslint and changing code style

Emulebest:
@shlomi-nx , @joshua-berry-ntnx please take a look at this and see the rules/changes you don't like. There is a lot of tedious work to be done here


papiea-engine/src/diff_resolver_main.ts
```diff
@@ -21,40 +21,66 @@ declare var process: {
...
+        watchlist,
+        watchlistDb,
+        specDb,
+        statusDb,
+        providerDb,
+        differ,
+        intentfulContext,
+        logger,
+        batchSize,
+    )
```
shlomi-nx, line 66: Do we have to have this line-abuse lint rule?

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -15,307 +15,445 @@ import {
...
-        this.validator = validator;
+    constructor(
+        logger: Logger,
+        status_db: Status_DB,
+        spec_db: Spec_DB,
+        provider_db: Provider_DB,
+        intentful_task_db: IntentfulTask_DB,
+        authorizer: Authorizer,
+        validator: Validator,
+        intentfulCtx: IntentfulContext,
```
shlomi-nx, line 49: Also here, why not two lines until 80 cols?

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -15,307 +15,445 @@ import {
...
+        sortParams?: SortParams,
+    ): Promise<Partial<IntentfulTask>[]> {
         const intentful_tasks = await this.intentful_task_db.list_tasks(fields, sortParams)
-        const entities = await this.spec_db.get_specs_by_ref(intentful_tasks.map(task => task.entity_ref))
-        const filteredRes = await this.authorizer.filter(user, entities, Action.Update, x => { return { "metadata": x[0] } });
+        const entities = await this.spec_db.get_specs_by_ref(
+            intentful_tasks.map(task => task.entity_ref),
+        )
+        const filteredRes = await this.authorizer.filter(
+            user, entities, Action.Update, x => { return { metadata: x[0] } },
```
shlomi-nx, line 82: Why the trailing `,` here? 

Also, this probably does not need to `return` and wrapping `{}`

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -15,307 +15,445 @@ import {
...
                 throw new Error("Uuid is not provided, but supposed to be since validation pattern is specified")
             }
         } else {
             const result = await this.get_existing_entities(provider, request_metadata.uuid)
             if (result.length !== 0) {
                 const [metadata, spec, status] = result
-                throw new ConflictingEntityError("An entity with this uuid already exists", metadata, spec, status)
+                throw new ConflictingEntityError(
+                    "An entity with this uuid already exists", metadata, spec, status,
+                )
```
shlomi-nx, line 114: Whats the purpose of this? Lets not just make new lines everywhere..

I see lots more of this kind of newline-abuse, wont comment more, but don't take it as if I am ok with them :|

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -15,307 +15,445 @@ import {
...
+            )
             if (result_spec.length !== 0 || result_status.length !== 0) {
                 return [result_spec[0][0], result_spec[0][1], result_status[0][1]]
-            } else {
-                return []
             }
+            return []
         } catch (e) {
             // Hiding details of the error for security reasons
             // since it is not supposed to occur under normal circumstances
```
shlomi-nx, line 459: Just to be clear, by this point we moved from 321 lines to 459, thats 138 newlines. This means a ~43% increase of newlines, most of which have no purpose and hinder readability in my opinion.

papiea-engine/src/diff_resolver_main.ts
Emulebest: Line length is 120 symbols, the general rule I use is if with one indentation parameter list is still too big, then I indent each parameter to a new line.

That's what I use, if you feel like it hurts readability then let's get rid of it

papiea-engine/src/entity/entity_api_impl.ts
Emulebest: Couldn't find a lint rule for one liners lambdas.

Trailing commas might be redundant IMO, what do you think?

papiea-engine/src/entity/entity_api_impl.ts
Emulebest: The new line above is necessary since 120 limit is exceeded without it

papiea-engine/src/entity/entity_api_impl.ts
joshua-berry-ntnx: I tend to like them because it helps to avoid polluting diffs when you are adding/removing items in a list. Basically they make `git blame` much more useful for trivial extra effort.

papiea-engine/src/diff_resolver_main.ts
joshua-berry-ntnx: I agree one simple identifier per line is a bit excessive; I don't mind it as much though if there's more going on on some of the lines (e.g. some type definition with type parameters or similar)

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -15,307 +15,445 @@ import {
...
+        input: any,
+    ): Promise<any> {
+        const provider = await this.get_provider(prefix, version)
+        const kind = this.providerDb.find_kind(provider, kind_name)
+        const entity_spec: [Metadata, Spec] = await this.get_entity_spec(
+            user, kind_name, entity_uuid,
+        )
+        const entity_status: [Metadata, Status] = await this.get_entity_status(
+            user, kind_name, entity_uuid,
+        )
```
joshua-berry-ntnx, line 230: I don't think putting single closing parentheses for function calls/expressions on their own line is really necessary. But I'm fine with breaking like this if there's something after the closing paren, e.g.

```js
const foo = bar(
    user, kind_name, entity_uuid, pszHungarianNotation, 0, /* reserved */, NULL /* reserved2 */,
    (void *)0 /* reserved3 */, FLAGS_NOFLAGS,
) + 128
```

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -15,307 +15,445 @@ import {
...
-                    headers: user
-                });
-            this.validator.validate(data, Object.values(procedure.result)[0], schemas,
-                provider.allowExtraProps, Object.keys(procedure.argument)[0], procedure_name);
-            return data;
+                    input,
+                },
+                {
+                    headers: user,
+                })
```
joshua-berry-ntnx, line 257: I'm similarly OK collapsing small objects like this onto one line:

```js
const {data} = axios.post(
    procedure.procedure_callback,
    {
        // big object
    },
    {littleObject: 'hi'})
```

papiea-engine/src/main.ts
```diff
@@ -35,67 +34,97 @@ declare var process: {
...
-    app.use(express.json());
-    app.use(auditLogger.middleware());
-    const mongoConnection: MongoConnection = new MongoConnection(mongoUrl, mongoDb);
-    await mongoConnection.connect();
+    const app = express()
+    app.use(cookieParser())
+    app.use(express.json())
+    app.use(auditLogger.middleware())
+    const mongoConnection: MongoConnection = new MongoConnection(mongoUrl, mongoDb)
+    await mongoConnection.connect()
```
joshua-berry-ntnx, line 59: My preference is to keep the semicolons (although I don't think we need to proactively add them); I've run into one too many ambiguous cases in the TypeScript grammar where a missing semicolon changed the meaning of the code and caused either compile-time or runtime issues.

Here is one example:

```js
const small_fn = (x) => x.toString()

(function () {
    // do some startup stuff which doesn't pollute the module namespace
})();
```

Here's another:

```js
async function foo() {
    openURL("http://whatever").catch(console.log)
    (await createBookmark("http://asdf")).save()
}
```

Basically any time you have a line which begins with `(`, you're going to have a bad time. I think there are some other cases which are escaping me at the moment (those are the two I've hit most frequently).

papiea-engine/src/entity/entity_api_impl.ts
shlomi-nx: arrrrggggg

papiea-engine/src/diff_resolver_main.ts
shlomi-nx: Can we relax the `1 symbol per line` thing. Before you know it you need to open 4 panes on one source file to see a few definitions all at once.

papiea-engine/src/entity/entity_api_impl.ts
shlomi-nx: I don't particularly mind trailing `,`. As a matter of fact, in some languages (lisps), a comma is treated as a whitespace by the parser. I agree to that notion.

papiea-engine/src/entity/entity_api_impl.ts
shlomi-nx: Agree to that too, no purpose in putting single closing paren in a new line.

papiea-engine/src/entity/entity_api_impl.ts
shlomi-nx: Agree with this too. so much more readable!

papiea-engine/src/main.ts
shlomi-nx: I personally don't like the old `;`.

I agree that some times ambiguities my come up, but forcing `;` on every line for those strange corner cases seems too much to me. I'd vote for avoiding them.

joshua-berry-ntnx: I don't like `;`s either, and in a perfect world, JS/TS wouldn't have them at all apart from exceptional cases where you want multiple expressions on a single line.

However, as someone who has lost hours to this ambiguity just in the last few months, I don't buy most of what's in the article—IME they are so inexpensive as to be indistinguishable from free, and this trivial cost is worth paying to avoid the introduction of bugs and time wasted tracking down strange compiler issues due to missing semicolons.

That said, I'm fine with omitting `;`s if we can get `tsc` itself to do something like `no-unexpected-multiline: 'error'`, which gets the compiler to catch those corner cases for you. I'd also be fine with relying on the linter to do it, but only if the linter is a mandatory part of every single build (which I think we said we didn't want to do, but maybe I'm wrong?).

shlomi-nx: I am fine with adding linter as a step in our cci job. However, we should probably use a linter that can automatically fix (most) errors (spacing, long lines etc). I do not want to have a cci job fails because of a whitespace. On ambiguous problems, no problems if either tsc or linter errors out.

Emulebest: I can put a prettier in the mix but I couldn't find a way to tie it down to our tslint rules so they are going to have 2 different configurations that we need to separately adjust if needed

shlomi-nx: Well, from https://palantir.github.io/tslint/
> :warning: TSLint has been deprecated as of 2019. Please see this issue for more details: Roadmap: TSLint → ESLint. typescript-eslint is now your best option for linting TypeScript.

(see also https://medium.com/palantir/tslint-in-2019-1a144c2317a9)

Eslint [supports auto-fixing](https://eslint.org/docs/user-guide/command-line-interface#--fix), and has a [typescript variant](https://github.com/typescript-eslint/typescript-eslint)

Emulebest:
Closing in favor of using eslint for typescript and rethinking the code style and changes

PR #431, opened 2020-06-23
Truncate spec|status-val postfix

### LINKED ISSUE #421: Rename `spec-val` and `status-val` of the Diff
We can simply drop the `-val` suffix. It could be retrofitted in the part which converts between clojure and js data-structures (no need to modify the differ logic itself)

Emulebest:
resolves #421


papiea-engine/__tests__/task_tests/intentful_task.test.ts
```diff
@@ -136,8 +136,8 @@ describe("Intentful Task tests", () => {
             }
         })
         expect(task.status).toEqual(IntentfulStatus.Pending)
-        expect(task.diffs[0].diff_fields[0]["spec-val"][0]).toEqual(20)
-        expect(task.diffs[0].diff_fields[0]["status-val"][0]).toEqual(null)
+        expect(task.diffs[0].diff_fields[0]["spec"][0]).toEqual(20)
+        expect(task.diffs[0].diff_fields[0]["status"][0]).toEqual(null)
```
shlomiv, line 140: You could change one of these to `.spec` notation to make sure it works too

papiea-engine/papiea-lib-clj/src/papiea-lib-clj/core.cljs
```diff
@@ -1,6 +1,7 @@
 (ns papiea-lib-clj.core
   (:require [cljs.nodejs :as nodejs]
-            [instaparse.core :as insta]))
+            [instaparse.core :as insta])
+  (:use [clojure.set :only (rename-keys)]))
```
shlomiv, line 4: Don't use `use`, instead do `:refer [rename-keys]` in `:require`

shlomiv:
Looks good, small comments

papiea-engine/papiea-lib-clj/src/papiea-lib-clj/core.cljs
Emulebest: I will change that, just out of curiosity why do we not use `use`?

PR #433, opened 2020-06-24
Better procedure description

### LINKED ISSUE #430: Support full documentation of Papiea procedures in Swagger
Right now, Papiea procedure definitions can only specify Swagger for the arguments and return values. There is no way for providers to include the complete documentation for a procedure, like error returns or a procedure summary/description, so the Swagger documentation generated by Papiea for procedures is incomplete.

It would be nice to have a way for a provider to specify full documentation for a procedure. At minimum this would include summary and description, and should probably also include error return codes/values.

Emulebest:
resolves #430


Emulebest:
The error from the test example in the api-docs look as follows:
image

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -415,7 +420,9 @@ export class Kind_Builder {
                      strategy: Procedural_Execution_Strategy,
                      input_desc: any,
                      output_desc: any,
-                     handler: (ctx: ProceduralCtx_Interface, entity: Entity, input: any) => Promise<any>): Kind_Builder {
+                     handler: (ctx: ProceduralCtx_Interface, entity: Entity, input: any) => Promise<any>,
+                     description?: string,
+                     error_descriptions?: ErrorDescriptions): Kind_Builder {
```
joshua-berry-ntnx, line 425: We have quite a few parameters for this function now. I'm thinking it might be easier to use if we collapsed input/output/description/errors into a single object which is a Swagger spec. (I think this also avoids/reduces the need for pre-processing of error descriptions?)

papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -521,7 +530,9 @@ export class Kind_Builder {
                    strategy: Procedural_Execution_Strategy,
                    input_desc: any,
                    output_desc: any,
-                   handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<any>): Kind_Builder {
+                   handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<any>,
+                   description?: string,
+                   error_descriptions?: ErrorDescriptions): Kind_Builder {
```
joshua-berry-ntnx, line 535: Same as above

papiea-engine/src/api_docs/api_docs_generator.ts
```diff
@@ -400,9 +401,56 @@ export default class ApiDocsGenerator {
...
+                        }
+                    }
+                }
+            }
+        }
+        if (sig.error_descriptions) {
+            for (let [key, val] of Object.entries(sig.error_descriptions)) {
+                if (proc_def.responses[key] === undefined) {
+                    proc_def.responses[key] = get_error_response(val.description, val.structure)
+                }
```
joshua-berry-ntnx, line 445: Should we complain if we are provided an error description which conflicts with an existing response definition?

papiea-engine/src/api_docs/api_docs_generator.ts
```diff
@@ -400,9 +401,56 @@ export default class ApiDocsGenerator {
...
+                                        "code",
+                                        "message"
+                                    ],
+                                    "properties": {
+                                        "errors": {
+                                            "type": "array",
+                                            "items": structure
+                                        },
+                                        "code": {
+                                            "type": "integer"
```
joshua-berry-ntnx, line 428: Is this conforming to HTTP or something else? If it's just our arbitrary thing, I'd prefer a string (which could simply be some well-known identifier that's easy to compare against and easy to read). (And I'd suggest calling it `type` or something similar to distinguish from `message` while avoiding a name suggesting it's numeric)

papiea-sdk/src/provider_sdk/typescript_sdk.ts
Emulebest: There is a https://github.com/nutanix/papiea/issues/427 issue, I'll reduce the number of parameters there. I don't think that a single object is a good representation because Swagger API has its own ruling for where to put those and I feel like what we have is quite clear

papiea-engine/src/api_docs/api_docs_generator.ts
Emulebest: Ideally, the error code could be restricted to only specific ranges (4xx-, 5xx-, etc)

papiea-engine/src/api_docs/api_docs_generator.ts
Emulebest: This part should be updated as we now already have `type` string in response

PR #435, opened 2020-06-25
Fix url escaping

### LINKED ISSUE #434: SDK 404's when engine POSTs to intent handler with array SFS (=> names/SFSes are improperly escaped)
I'm using the TypeScript SDK version 1.0.86, registering an intent handler like so:

```ts
const subnet = app.new_kind({subnet: ...});
subnet.on('used_ips.+{ip}', null, Subnet.on_ip_updated);
```

When the engine tries to invoke this intent handler on the provider, it 404's like so (collected with the help of additional debugging logs added to the engine):

```
papiea_1  |     request: ClientRequest {
...
papiea_1  |       _header: 'POST /subnet/used_ips.+%7Bip%7D HTTP/1.1\r\n' +
papiea_1  |         'Accept: application/json, text/plain, */*\r\n' +
papiea_1  |         'Content-Type: application/json;charset=utf-8\r\n' +
papiea_1  |         'User-Agent: axios/0.18.1\r\n' +
papiea_1  |         'Content-Length: 720\r\n' +
papiea_1  |         'Host: bmaas:9198\r\n' +
papiea_1  |         'Connection: close\r\n' +
papiea_1  |         '\r\n',
...
papiea_1  |     data: '<!DOCTYPE html>\n' +
papiea_1  |       '<html lang="en">\n' +
papiea_1  |       '<head>\n' +
papiea_1  |       '<meta charset="utf-8">\n' +
papiea_1  |       '<title>Error</title>\n' +
papiea_1  |       '</head>\n' +
papiea_1  |       '<body>\n' +
papiea_1  |       '<pre>Cannot POST /subnet/used_ips.+%7Bip%7D</pre>\n' +
papiea_1  |       '</body>\n' +
papiea_1  |       '</html>\n',
...
papiea_1  |     status: 404,
papiea_1  |     statusText: 'Not Found'
```

I suspect the problem might be the `+` character in the URL is not properly escaped, as a url-encoded `+` is usually just a space. Or it could be the problem is that there's _too much_ escaping going on, because I'm looking at the SDK code and I don't see any escaping being done when registering the routes for receiving intent-handler calls from the engine (see typescript_sdk.ts):

```ts
class Provider_Server_Manager {
    ...
    procedure_callback_url(procedure_name: string, kind?: string): string {
        if (kind !== undefined) {
            return `http://${ this.public_host }:${ this.public_port }/${ kind }/${ procedure_name }`
        } else {
            return `http://${ this.public_host }:${ this.public_port }/${ procedure_name }`
        }
    }
    ...
}

class Kind_Builder {
    ...
    on(sfs_signature: string, ...) {
        ...
        this.server_manager.register_handler(`/${this.kind.name}/${sfs_signature}`, ...)
        ...
    }
}
```

We should take care to make sure procedure names and SFS signatures are properly URL-encoded on both the engine and SDK sides. Otherwise you could do silly things like register procedures with a name of `/.../:foo/bar`, and this will almost certainly not do what anyone expects.

Emulebest:
resolves #434


papiea-sdk/src/provider_sdk/typescript_sdk.ts
```diff
@@ -341,18 +341,25 @@ class Provider_Server_Manager {
         this.app.use(express.json())
     }
 
+    private escape_sfs_path(path: string) {
+        const replaced_once = path.replace(/[*+?()]/g, '\\$&') // $& means the whole matched string
```
joshua-berry-ntnx, line 345: If this is trying to quote a regex, I think you might need a few more things (not sure which of these Express actually pays attention to):

`path.replace(/[.*+?|(){}]|\[|\]|\\/g, '\\$&')`

Emulebest: Good catch, though I've specifically went through Express docs and escaped only symbols that it uses, all the others shouldn't be escaped since it breaks the routing

PR #440, opened 2020-07-02
Filter & List Papiea-Client generator implementation

### LINKED ISSUE #429: SDK enhancements to filter/list entities
It would be really nice if `EntityCRUD` had a few more features for filtering/listing entities:

* A `list()` function which returns an async iterator over all entities (and incrementally fetches entities using the appropriate offset/limit query parameters)
* Update `filter()` to follow a similar pattern—return an async iterator which incrementally fetches matching entities, as this is what the vast majority of users will want.
* Take a parameter to `list()` and `filter()` which specifies the batch size—so for example, if you call `filter()` intending only to fetch 1 or 2 entities, you can give this hint to `filter()` to avoid fetching too many entities needlessly.

Additionally, for `filter()` it might be nice (but less critical) to have a similar API to what we have now to return (an estimate of) the total number of results without having to actually fetch all the results. Maybe `count()`?

I imagine this will be relevant for both TypeScript and Python SDKs eventually.

Emulebest:
resolves #429

### DIFF
--- a/papiea-client/src/entity_client.ts
+++ b/papiea-client/src/entity_client.ts
@@ -109,7 +109,6 @@ export async function filter_entity(provider: string, kind: string, version: str
 
 export async function iter_filter(url: string, data: any, config?: AxiosRequestConfig) {
     return (async function* iter(batch_size?: number, offset?: number): AsyncGenerator<any, undefined, any> {
-        console.log(offset)
         if (!batch_size) {
             batch_size = BATCH_SIZE
         }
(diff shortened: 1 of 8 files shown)


papiea-client/src/entity_client.ts
```diff
@@ -95,11 +97,33 @@ export interface FilterResults {
...
+
 export async function filter_entity(provider: string, kind: string, version: string, filter: any, papiea_url: string, s2skey: string): Promise<FilterResults> {
-    const res = await make_request(axios.post, `${ papiea_url }/services/${ provider }/${ version }/${ kind }/filter/`, filter, { headers: { "Authorization": `Bearer ${ s2skey }` } });
+    const res = await make_request(axios.post, `${ papiea_url }/services/${ provider }/${ version }/${ kind }/filter`, filter, { headers: { "Authorization": `Bearer ${ s2skey }` } });
     return res.data
 }
 
+export async function iter_filter(url: string, data: any, config?: AxiosRequestConfig) {
+    return (async function* iter(batch_size?: number, offset?: number): AsyncGenerator<any, undefined, any> {
+        console.log(offset)
```
shlomi-nx, line 112: log with no text

Emulebest: Sorry a debug leftover

PR #472, opened 2020-07-24
Added tests & added error message in case of failed status watcher

### LINKED ISSUE #469: Two conflicting concurrent updates both appear to succeed
If one attempts two concurrent PUTs against the same entity with the same spec and spec_version, I would expect one of those requests to fail with an HTTP 409 Conflict return code.

However, neither update appears to be failing—both PUTs return HTTP 200s, and the entity adopts the new spec_version. In looking at the Papiea log output, it seems like there IS actually some conflict handling happening—I notice that the PUT which is serviced second returns a watcher with a `Failed` status, but there is no reason/additional info supplied to indicate there's a conflict. For example:

```
papiea_1  | audit        2020-07-23T00:43:46.443Z -- {
...
papiea_1  |   method: 'PUT',
papiea_1  |   request_body: {
papiea_1  |     metadata: { spec_version: 1 },
papiea_1  |     spec: {
papiea_1  |       l33t_hax: {
...
papiea_1  |   },
papiea_1  |   request_ip: '::ffff:172.22.0.4',
papiea_1  |   response_body: {
papiea_1  |     watcher: {
papiea_1  |       _id: '5f18dd4209f15e0032a75533',
papiea_1  |       created_at: '2020-07-23T00:43:46.424Z',
papiea_1  |       diffs: [
...
papiea_1  |           ],
papiea_1  |           intentful_signature: {
...
papiea_1  |           },
papiea_1  |           kind: 'machine'
papiea_1  |         }
papiea_1  |       ],
papiea_1  |       entity_ref: { kind: 'machine', uuid: '28b42110-0744-4143-8e80-60d30717d236' },
papiea_1  |       spec_version: 2,
papiea_1  |       status: 'Failed',
papiea_1  |       times_failed: 0,
papiea_1  |       user: {
...
papiea_1  |       },
papiea_1  |       uuid: 'c61caf26-42a7-47f6-b0c7-c7b7d1bdc862'
papiea_1  |     }
papiea_1  |   },
papiea_1  |   status_code: 200,
papiea_1  |   url: '/services/bmaas/0.0.1/machine/28b42110-0744-4143-8e80-60d30717d236',
papiea_1  |   user: {
...
papiea_1  |   }
papiea_1  | }
```

(The first PUT to be serviced always returns a watcher with `Pending` status.)

This ONLY seems to happen with two PUTs that happen concurrently/very close together in time—when attempting two PUTs serially in the UI, the second fails with HTTP 409 as it should. So it seems there is some sort of race in the Papiea engine.

Emulebest:
resolves #469 

@joshua-berry-ntnx I couldn't find the race condition described in the issue and added a few tests for concurrent and serial updates. A few words though:
1) `Differ` entities always return Watcher as a response to any PUT request (Failed one in case of CAS problem)
2) All the other entities return `409 Conflicting Entity` Error in case of CAS problem

Thus for `Differ` entities I added error message in case of CAS fail to signify what is the problem exactly.

### DIFF
--- a/papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
+++ b/papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
@@ -718,6 +718,138 @@ describe("Intentful Workflow tests", () => {
         }
     })
 
+    test("Two concurrent updates with the same spec version should fail", async () => {
+        expect.assertions(4);
+        const sdk = ProviderSdk.create_provider(papieaUrl, adminKey, server_config.host, server_config.port);
+        try {
+            first_provider_prefix = "location_provider_intentful_2_concurrent_cas"
+            const location = sdk.new_kind(locationDataDescription);
+            sdk.version(provider_version);
+            sdk.prefix(first_provider_prefix);
+            location.on("x", async (ctx, entity, input) => {
+                await providerApiAdmin.patch(`/${sdk.provider.prefix}/${sdk.provider.version}/update_status`, {
+                    context: "some context",
+                    entity_ref: {
+                        uuid: metadata.uuid,
+                        kind: kind_name
+                    },
+                    status: { x: entity.spec.x }
+                })
+            })
+            location.on_create(async (ctx, entity) => {
+                const { metadata, spec } = entity
+                await ctx.update_status(metadata!, spec!)
+            })
+            await sdk.register();
+            const kind_name = sdk.provider.kinds[0].name;
+            const { data: { metadata, spec } } = await entityApi.post(`${ sdk.entity_url }/${ sdk.provider.prefix }/${ sdk.provider.version }/${ kind_name }`, {
+                spec: {
+                    x: 10,
+                    y: 11
+                }
+            })
+            first_provider_to_delete_entites.push(metadata)
+            await timeout(5000)
+            const promises = []
+            for (let i = 0; i < 2; i++) {
+                promises.push(entityApi.put(`/${ sdk.provider.prefix }/${ sdk.provider.version }/${ kind_name }/${ metadata.uuid }`, {
+                    spec: {
+                        x: 20,
+                        y: 11
+                    },
+                    metadata: {
+                        spec_version: 1
+                    }
+                }))
+            }
+            const results = await Promise.all(promises)
+            try {
+                const first_watcher_result = await entityApi.get(`/intent_watcher/${ results[0].data.watcher.uuid }`)
+                const second_watcher_result = await entityApi.get(`/intent_watcher/${ results[1].data.watcher.uuid }`)
+                if (second_watcher_result.data.status === IntentfulStatus.Failed && first_watcher_result.data.status === IntentfulStatus.Pending) {
+                    expect(second_watcher_result.data.status).toBe(IntentfulStatus.Failed)
+                    expect(second_watcher_result.data.times_failed).toEqual(1)
+                    expect(second_watcher_result.data.last_handler_error).toEqual("Spec with this version already exists")
+                    expect(first_watcher_result.data.status).toBe(IntentfulStatus.Pending)
+                    return
+                }
+            } catch (e) {
+                console.log(`Couldn't get entity: ${e}`)
+            }
+        } finally {
+            sdk.server.close();
+        }
+    })
+
+    test("Two serial updates with the same spec version should fail", async () => {
+        expect.assertions(3);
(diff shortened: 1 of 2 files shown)


papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
```diff
@@ -718,6 +718,138 @@ describe("Intentful Workflow tests", () => {
...
+                    spec: {
+                        x: 20,
+                        y: 11
+                    },
+                    metadata: {
+                        spec_version: 1
+                    }
+                }))
+            }
+            const results = await Promise.all(promises)
```
joshua-berry-ntnx, line 765: This doesn't seem right to me—at least one update should have failed. I would expect this to throw a `ConflictingEntityError`.

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
```diff
@@ -718,6 +718,138 @@ describe("Intentful Workflow tests", () => {
...
+                const second_watcher_result = await entityApi.get(`/intent_watcher/${ results[1].data.watcher.uuid }`)
+                if (second_watcher_result.data.status === IntentfulStatus.Failed && first_watcher_result.data.status === IntentfulStatus.Pending) {
+                    expect(second_watcher_result.data.status).toBe(IntentfulStatus.Failed)
+                    expect(second_watcher_result.data.times_failed).toEqual(1)
+                    expect(second_watcher_result.data.last_handler_error).toEqual("Spec with this version already exists")
+                    expect(first_watcher_result.data.status).toBe(IntentfulStatus.Pending)
+                    return
+                }
+            } catch (e) {
+                console.log(`Couldn't get entity: ${e}`)
```
joshua-berry-ntnx, line 777: Disregarding my previous comment, should we not fail the test at this point?

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
```diff
@@ -718,6 +718,138 @@ describe("Intentful Workflow tests", () => {
...
+                const second_watcher_result = await entityApi.get(`/intent_watcher/${ second_result.data.watcher.uuid }`)
+                if (second_watcher_result.data.status === IntentfulStatus.Failed &&
+                    (first_watcher_result.data.status === IntentfulStatus.Pending || first_watcher_result.data.status === IntentfulStatus.Completed_Successfully)) {
+                    expect(second_watcher_result.data.status).toBe(IntentfulStatus.Failed)
+                    expect(second_watcher_result.data.times_failed).toEqual(1)
+                    expect(second_watcher_result.data.last_handler_error).toEqual("Spec with this version already exists")
+                    return
+                }
+            } catch (e) {
+                console.log(`Couldn't get entity: ${e}`)
```
joshua-berry-ntnx, line 846: Shouldn't this cause a test failure?

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
Emulebest: Please see the comments on the PR on the expected behavior I implemented. TLDR; All the differ entities return a watcher (even if a Failed one) and the error if present is in the watcher field

joshua-berry-ntnx: Yeah sorry, I should have written a clearer comment—I get that's the intention here, I just don't think it's the right thing to do.

1. There is more than one way for clients to observe entity status changes—they can use the diff watcher that's returned, or they can simply poll an entity with get(). If they do the latter (which is what bmaas always does, for example), there is no indication the update failed.
2. More generally, as with any other return-value based error reporting, clients can easily ignore the return value and have no idea an error occurred. This might be idiomatic in other languages but it is not idiomatic in JavaScript and is likely to be surprising to users of the SDK, especially when other types of errors (e.g. Papiea server isn't reachable, etc.) result in exceptions.
3. To make matters worse, only _some_ entity types report conflicts through diff-watchers. Others actually will throw `ConflictingEntityError`. We should pick one behavior and stick with it rather than requiring users to understand the type of entity they're dealing with. Users should not need to know or care whether they are dealing with a differ entity or some other type of entity.

shlomi-nx: I think thats a good point.. Let's throw a conflict error when unable to CAS on all entity types.

PR #473, opened 2020-07-27
Spec version range in graveyard implemented

### LINKED ISSUE #471: spec_version should not be reused for different generations of the same entity
Per https://github.com/nutanix/papiea/issues/455#issuecomment-662803382, updates to an entity which have been deleted and subsequently recreated should fail. From the original comment:

> One thing we need to think about: spec_version. We should probably ensure spec_version is somehow unique or different between previous and new versions of a particular entity. Otherwise you could wind up in pathological cases like this:
> 
> 1. Client A fetches info on an entity, which has spec_version 1, then gets delayed, swapped out, decides to take a nap, etc.
> 2. Client B deletes the entity
> 3. Client C creates the entity again, which gets spec_version 1.
> 4. Client A wakes up and tries updating the entity with spec_version 1. The update succeeds, even though the entity has been deleted and recreated.
> 
> So I think we may want to consider persisting the spec_version even after the entity is gone, at least for a reasonable amount of time (and make deletion be a spec_version increment).

Emulebest:
resolves #471

### DIFF
--- a/papiea-engine/__tests__/entity_tests/entity_api.test.ts
+++ b/papiea-engine/__tests__/entity_tests/entity_api.test.ts
@@ -494,7 +494,7 @@ describe("Entity API tests", () => {
     });
 
     test("Recreate deleted entity with the same spec version should fail", async () => {
-        expect.assertions(1);
+        expect.assertions(2);
         const id = uuid()
         const { data: { metadata, spec } } = await entityApi.post(`/${ providerPrefix }/${ providerVersion }/${ kind_name }`, {
             spec: {
@@ -519,6 +519,7 @@ describe("Entity API tests", () => {
             });
         } catch (e) {
             expect(e.response.status).toEqual(409)
+            expect(e.response.data.error.message).toEqual(`Deleted entity with this uuid and spec version exists: uuid - ${id}, maximum current spec version - 2`)
         }
     });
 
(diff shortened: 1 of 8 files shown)


papiea-engine/__tests__/entity_tests/entity_api.test.ts
```diff
@@ -493,6 +493,60 @@ describe("Entity API tests", () => {
...
+                    x: 10,
+                    y: 11
+                },
+                metadata: {
+                    uuid: id,
+                    spec_version: 1
+                }
+            });
+        } catch (e) {
+            expect(e.response.status).toEqual(409)
```
shlomi-nx, line 521: I am just worried a client may be surprised by this. They will try to GET and see there is nothing there, and then they will try to create and see a "conflict". I see that you put in a good error message, is it a payload in the response?

papiea-engine/src/intentful_core/intentful_strategies/intentful_strategy_interface.ts
```diff
@@ -20,13 +21,25 @@ export abstract class IntentfulStrategy {
         this.graveyardDb = graveyardDb
     }
 
+    protected async check_spec_version(metadata: Metadata, spec_version: number, spec: Spec) {
+        const exists = await this.graveyardDb.check_spec_version_exists(metadata, spec_version)
+        if (exists) {
+            metadata.spec_version = spec_version
+            throw new ConflictingEntityError(
+                "Deleted entity with this uuid and spec version exists", metadata, spec)
```
shlomi-nx, line 29: Perhaps it would be convenient to also include in the error string the *current* maximum spec version, for users to see.

PR #475, opened 2020-07-28
Conflicting entities now always throw the exception

Emulebest:
@shlomi-nx now that all the entities throw, what is the use for `Failed` status on IntentWatcher? I've left it just in case we would like to have history of intent changes illustrated by watchers but still


shlomi-nx:
Hmm, not sure at the moment. Lets leave it for now.

PR #479, opened 2020-07-30
Upgrade version scripts and README version section

Emulebest:


switch-registry.sh
```diff
@@ -0,0 +1,14 @@
...
+  npm config set registry https://nutanix.jfrog.io/nutanix/api/npm/npm-virtual/
+  yarn config set registry https://nutanix.jfrog.io/nutanix/api/npm/npm-virtual/
+  echo 'Registry set to https://nutanix.jfrog.io/nutanix/api/npm/npm-virtual/'
+  ;;
+"default")
+  npm config set registry https://registry.npmjs.org/
+  yarn config set registry https://registry.npmjs.org/
+  echo 'Registry set to https://registry.npmjs.org/'
+  ;;
+esac
```
shlomiv, line 14: Should we also include some standard `docker-hub` like registry for the docker file? (of course after we optimize it from its current gigantic size..)

README.md
```diff
@@ -9,6 +9,12 @@ between intended state and real world state.
 
 Please see [Papiea's design document](https://nutanix.github.io/papiea/Papiea-design.html)
 
+## Papiea versions:
+
+| Papiea-engine docker version  | Typescript SDK/Client version | Python SDK/Client commit |
+| ------------- | ------------- | ------------- |
+| 1298  | 1.0.92  | 581a46b95be2485a403079d2f82d09f4b1cb0411 |
+
```
shlomiv, line 17: Actually, could we make this a vertical table instead?

| Component | Version |
|------------|------|
|Engine (docker)|1298|
|Client/SDK (typescript)|1.0.92|
|Client/SDK (python)| 581a46b95be2485a403079d2f82d09f4b1cb0411|

switch-registry.sh
Emulebest: That would be a good idea!

README.md
Emulebest: Sure

upgrade-versions.sh
```diff
@@ -0,0 +1,99 @@
...
+
+if [ "$YARN_REGISTRY" != "registry: 'https://nutanix.jfrog.io/nutanix/api/npm/npm-virtual/'," ]
+then
+  echo "Your yarn registry is not set to nutanix jfrog!"
+  exit 1
+fi
+
+collect_args "$@"
+
+if [[ -z "${CIRCLE_BUILD_NUM}" ]]; then
```
joshua-berry-ntnx, line 47: This is a bash-ism, but the extension is `.sh` (and there is no shebang line). If you intend this to be a bash script, please add a shebang line reflecting that (and we should be consistent in our use of `test`, `[` and `[[`).

upgrade-versions.sh
```diff
@@ -0,0 +1,99 @@
+# This is a script that updates typescript SDK, prints out
+# Typescript SDK version, Python commit hash
+# You use these to update README.md section 'Papiea versions'
```
joshua-berry-ntnx, line 3: `set -e` somewhere before the first actual line of the script (since you have places later in the script where you assume that commands succeed)

Issue #492, opened 2020-08-31
Refactor kind creation factory


Emulebest:
Ideally we would utilize a usability pattern of the following:
```
const kind = getKind(spec_only|basic|differ)
                                   .withDescription(array|location|cluster)
                                   .withStatusOnlyField()
                                   .withNullableField()
```
The example of this pattern could be seen here https://github.com/nutanix/papiea/blob/3a381a7a9f9253acc8fd00c3d305b68c3a6d2da5/papiea-engine/__tests__/test_data_factory.ts#L186

Emulebest:
Introduce the human-readable form to interpret the kind/description

PR #494, opened 2020-09-02
Kind description builder for tests fixtures

NikitaChernenkoNutanix:
issues #492


papiea-core/src/core.ts
```diff
@@ -144,6 +144,10 @@ export enum IntentfulBehaviour {
     Differ = "differ"
 }
 
+export enum FieldBehavior {
+    StatusOnly = "status-only"
+}
```
Emulebest, line 149: This is good!

papiea-engine/__tests__/test_data_factory.ts
```diff
@@ -33,180 +34,25 @@ export function loadYamlFromTestFactoryDir(relativePath: string): any {
...
-export function getSpecOnlyKindDescription(): Data_Description {
-    return getLocationDataDescription()
-}
-
-export function getSpecOnlyKindDescriptionWithSpecOnlyFields(): Data_Description {
-    const description: any = getLocationDataDescription();
-    description[Object.keys(description)[0]].properties.y["x-papiea"] = "spec-only"
-    return description
-}
+export function getKind(type: IntentfulBehaviour, desc?: Data_Description, signatures?: Intentful_Signature[]): Kind {
```
Emulebest, line 38: I wouldn't put signatures in here, since they are kind of close to procedures that we already have in ProviderBuilder, mind moving them there?

papiea-engine/__tests__/test_data_factory.ts
```diff
@@ -623,3 +469,105 @@ export class OAuth2Server {
...
+        this.withField(name, {"type": type || "number", "x-papiea": IntentfulBehaviour.SpecOnly})
+        return this
+    }
+
+    public withRequiredField(name: string) {
+        this.additionalRequiredFields.push(name)
+        return this
+    }
+
+    public withIntentfulBehaviour(behavior?: IntentfulBehaviour) {
```
Emulebest, line 562: I would make a default parameter here rather than checking if it is undefined in other places

papiea-engine/__tests__/test_data_factory.ts
```diff
@@ -623,3 +469,105 @@ export class OAuth2Server {
...
+        return this
+    }
+
+    public withStatusOnlyField(name?: string, type?: string) {
+        this.withField(name, {"type": type || "number", "x-papiea": FieldBehavior.StatusOnly})
+        return this
+    }
+
+    public withSpecOnlyField(name?: string, type?: string) {
+        this.withField(name, {"type": type || "number", "x-papiea": IntentfulBehaviour.SpecOnly})
```
Emulebest, line 553: Please use default params or operator `??` https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

papiea-engine/__tests__/test_data_factory.ts
```diff
@@ -623,3 +469,105 @@ export class OAuth2Server {
...
+    }
+
+    public withNullableFields() {
+        this.makeFeldsNullable = true
+        return this
+    }
+
+
+    public withField(name?: string, structure?: any) {
+        let key = name || "field_" + randomString(5)
```
Emulebest, line 542: Please use default params or operator ?? https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html

papiea-engine/__tests__/test_data_factory.ts
Emulebest, line 562: I would also suggest calling it `withBehavior` since there is too much Intentful in saying `new DescriptionBuilder().withIntentfulBehaviour(IntentfulBehaviour.SpecOnly)` when in reality it is not even intentful at all :). In hindsight, calling the enum `IntentfulBehaviour` was a bad idea from me, we should rename it all together with `IntentfulBehaviour.Differ` to something like `Behavior.Intentful` but that's for another issue

Emulebest (changes requested):
There are some small changes I'd hope to see. Thinking about it more, why do we even have `IntenfulBehaviour` and `x-papiea-entity` when the do the same things, maybe we should remove one of those, @shlomi-nx what do you think?

papiea-engine/__tests__/test_data_factory.ts
NikitaChernenkoNutanix: alright, let's have them in one class

papiea-engine/__tests__/test_data_factory.ts
NikitaChernenkoNutanix: for metadata we want it to be the nullable value of intentful behavior
https://github.com/nutanix/papiea/blob/kind-description-builder-for-tests-fixtures/papiea-engine/__tests__/test_data_factory.ts#L501

papiea-engine/__tests__/test_data_factory.ts
NikitaChernenkoNutanix: will do

papiea-engine/__tests__/test_data_factory.ts
NikitaChernenkoNutanix: decided to introduce KindBuilder

NikitaChernenkoNutanix: done

papiea-engine/__tests__/test_data_factory.ts
NikitaChernenkoNutanix: done

papiea-engine/__tests__/test_data_factory.ts
NikitaChernenkoNutanix: done

papiea-engine/__tests__/test_data_factory.ts
NikitaChernenkoNutanix: done

Emulebest: Okay

Emulebest (approved):
Looks much better, thanks!

Issue #485, opened 2020-08-05
Make a Changelog for Papiea


Emulebest:
For now let's have it automatically generated out of commits.

Emulebest:
Let's instead pull the changelog from issues but for now keep it simple (name and link)

Emulebest:
Keep in mind that we need optional manual descriptions for the major releases

NikitaChernenkoNutanix:
One of the possible implementations to have a well-structured changelog alongside with proper semantic versioning is to use this approach to name commits:
https://www.conventionalcommits.org/en/v1.0.0/

It will bring us an ability to correctly bump versions on merge to master (as currently, we keep updating patch version which is not according to semantic versioning)

As well as that will help to create a well-structured log out of the commits on merge to master. For example, there is a python lib that works well to fix such structure. https://pypi.org/project/python-changelog/. It expands a file named Changelog.md
that we will be able to correct manually on major releases.

NikitaChernenkoNutanix:
If the structure of conventional commits seems to tight, we can determine our own commit structure and set the rules we want to parse it to changelog. One of the solutions is https://pypi.org/project/gitchangelog/ that works with custom commit structure to create changelog.

However, it might not help us with versioning (there may be a parser of the custom structure for version bumping, needs further investigation).

NikitaChernenkoNutanix:
The last option is to directly print commit messages with the structure we have now in the changelog. it doesn't require a change of the way we write commits but may bring less value to our users.

joshua-berry-ntnx:
Nice, I didn't know about conventional commits. I do think it's going to be a bit hard for us to remember to follow though.

Did you also get a chance to look into generating a ChangeLog from issues/PRs? I don't remember if you were part of the discussion with Shlomi, Igor and I on this topic, but I think issues/PRs are generally going to be a bit higher-level and would make for more useful inputs for a ChangeLog—IMO we can start with something relatively basic like capturing issue/PR summaries and providing links back to each issue/PR, and see how that looks.

Emulebest:
Went through the docs, agree with @joshua-berry-ntnx. Moreover, tight commit structure could be a road block for potential new contributors to papiea, especially if we are going to use something custom

NikitaChernenkoNutanix:
then let's go with PR name and link to it (that basically means commit message title and the link to the PR) for now to keep it simple.

Emulebest:
closed by #500

PR #498, opened 2020-09-08
introduce appending changes to CHANGELOG.md when PR is merged to master

NikitaChernenkoNutanix:
issue #485 

appends message from the last commit when things are merged to master. (Assumed that the PR is squashed to 1 commit)

example of the addition to the CHANGELOG.md:
- Version 1.1: Upgrade version scripts and README version section (#479) PR url: https://github.com/nutanix/papiea/pull/479

### DIFF
--- a/upgrade-versions.sh
+++ b/upgrade-versions.sh
@@ -97,8 +97,8 @@ fi
 
 # append version with PR link to the CHANGELOG.md
 commit_message=$(git log -n 1 --format=%s)
-# match (#$number) in the commit message
-pr_number=$(echo $commit_message | grep -oP "\(\#\K\d+")
+# match #$number in the commit message
+pr_number=$(echo $commit_message | grep -E "#[1-9]+\b")
 
 changelog_message="- Version $typescript_version: $commit_message"
 # if there is a PR number in the commit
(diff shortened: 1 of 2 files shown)


upgrade-versions.sh
```diff
@@ -95,6 +95,18 @@ else
   " README.md
 fi
 
+# append version with PR link to the CHANGELOG.md
+commit_message=$(git log -n 1 --format=%s)
+# match (#$number) in the commit message
+pr_number=$(echo $commit_message | grep -oP "\(\#\K\d+")
```
joshua-berry-ntnx, line 101: `-P` does not work on non-Linux OSes. Try `grep -E` instead.

Also, I'd suggest something slightly different — look for `#1234` at word boundaries, not just inside parentheses — this will catch things like `fixed #1234` where there is no punctuation to look at.

upgrade-versions.sh
```diff
@@ -95,6 +95,18 @@ else
...
+
+changelog_message="- Version $typescript_version: $commit_message"
+# if there is a PR number in the commit
+if [[ "$pr_number" -ne "" ]]; then
+  changelog_message+=" PR url: $CIRCLE_REPOSITORY_URL/pull/$pr_number"
+fi
+echo $changelog_message >> CHANGELOG.md
```
joshua-berry-ntnx, line 108: Don't forget to `git add`

Emulebest (approved):
LGTM

PR #507, opened 2020-09-22
Exponential backoff implementation

### LINKED ISSUE #504: Provide exponential backoff for providers' intentful handlers

Emulebest:
resolves #504

### DIFF
--- a/papiea-engine/src/utils/utils.ts
+++ b/papiea-engine/src/utils/utils.ts
@@ -2,6 +2,8 @@ import { SortParams } from "../entity/entity_api_impl"
 import { ValidationError } from "../errors/validation_error"
 import { AxiosError } from "axios"
 
+export const DEFAULT_ENTROPY_FN = () => getRandomInt(10, 20)
+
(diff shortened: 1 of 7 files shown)


papiea-engine/src/utils/utils.ts
```diff
@@ -124,3 +124,25 @@ export function deepMerge(target: any, ...sources: any[]): any {
 
     return deepMerge(target, ...sources);
 }
+
+export function calculateBackoff(n: number, maximumBackoff: number, entropy: number) {
+    return Math.min(Math.pow(2, n) + entropy, maximumBackoff)
```
joshua-berry-ntnx, line 129: Consider making the entropy part of the exponent, so you compute backoff based on something like `Math.pow(2, n + entropy)` where `0 < entropy < 1`. This will give you a bit more variation at each stage of the retry, while still generally following an exponential curve.

(btw, let's also pick a better name for `n` -- I assume `n` is the number of retries, so maybe just call it `retries`)

joshua-berry-ntnx (approved):

PR #510, opened 2020-09-28
Added papiea customization via file config

### LINKED ISSUE #503: Provide more convenient customization to papiea engine
We need to provide more customization options (TTL for deleted entities, delay timings, strategies, etc.) to papiea. We should employ best practices from highly-customizable systems like JVM, docker.

Emulebest:
resolves #503


README.md
```diff
@@ -82,14 +82,55 @@ Now you can use your `${PAPIEA_INSTANCE_PUBLIC_IP}:3333` as PAPIEA_URL in benchm
...
-A set of these variables might be used to tweak the default papiea-engine configuration
-
-* `MONGO_HOST` - mongo host (default `mongo`)
-* `MONGO_PORT` - mongo port (default `27017`)
-* `PAPIEA_PUBLIC_URL` - external papiea address (default `"http://localhost:3000"`)
-* `HOT_RELOAD` - use nodemon to autoreload papiea on code changes (default `false`)
-* `DEBUG_LEVEL` - log level for [morgan](https://github.com/expressjs/morgan) logging (default `common`)
-* `PAPIEA_ADMIN_S2S_KEY` - papiea admin's s2s key
+Papiea uses configuration file to apply settings. Sample config file is located in `papiea-engine/papiea-config.json`.
+The configuration parameters are following:
```
joshua-berry-ntnx, line 86: This might be a style thing but can we use YAML for the config file instead? I know people tend to have strong opinions about YAML, but YAML allows for comments, and I think being able to leave comments in a config file is really really important (moreso than any stylistic objections).

To illustrate the point, your example below is not valid JSON. ;)

papiea-engine/src/utils/arg_parser.ts
```diff
@@ -0,0 +1,83 @@
...
+    // Deleted watcher persists in database for this amount of seconds
+    deleted_watcher_persist_time: number
+}
+
+const TRANSFORM_FN_MAP: { [key: string]: Function } = {
+    server_port: Number.parseInt,
+    entity_batch_size: Number.parseInt,
+    deleted_watcher_persist_time: Number.parseInt,
+    debug: (val: string) => val === "true"
+}
```
joshua-berry-ntnx, line 41: Very nice. :) Can I suggest clubbing this with the interface definition above and mapping the FN_MAP to the corresponding type? Something like:

```ts
const TRANSFORM_FN_MAP = {...} as const;
export type PapieaConfig = {
    [k in keyof (typeof TRANSFORM_FN_MAP)]: ReturnType<(typeof TRANSFORM_FN_MAP)[k]>,
};
```

Then you can avoid duplicating what are effectively two different type definitions--one for the transformations and one for the config object itself.

papiea-engine/src/utils/arg_parser.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    // Deleted watcher persists in database for this amount of seconds
+    deleted_watcher_persist_time: number
+}
+
+const TRANSFORM_FN_MAP: { [key: string]: Function } = {
+    server_port: Number.parseInt,
+    entity_batch_size: Number.parseInt,
+    deleted_watcher_persist_time: Number.parseInt,
+    debug: (val: string) => val === "true"
```
joshua-berry-ntnx, line 40: For these booleans, can I suggest checking for `not in [0, null, false, ""]` and their string equivalents instead? For environment variables, often people will do silly/stupid things like `PAPIEA_DEBUG=foo` and expect that because the variable is not the empty string, debug will be enabled (since many shell scripts are written this way).

papiea-engine/src/utils/arg_parser.ts
```diff
@@ -0,0 +1,83 @@
...
+export function getConfig(): PapieaConfig {
+    const config: PapieaConfig = JSON.parse(fs.readFileSync(PAPIEA_CONFIG_PATH, 'utf-8'))
+    const mapConfigToEnv: { [key: string]: string } = {}
+    Object.assign(mapConfigToEnv, config)
+    for (let key in mapConfigToEnv) {
+        mapConfigToEnv[key] = `PAPIEA_${key.toUpperCase()}`
+    }
+    // If there is an env variable prefixed with PAPIEA_ and has the same name
+    // as the config parameter - override config param with an env variable
+    for (let key in config) {
```
joshua-berry-ntnx, line 64: Check all possible environment variables (including defaults), not just those specified in config. Conceptually, you want something like:

```ts
let result = {};
Object.assign(result, defaults, config, env_vars);
```

So if something is set in defaults but not config, it can still be overridden.

joshua-berry-ntnx (changes requested):

README.md
Emulebest: I don't mind having a YAML definition

papiea-engine/src/utils/arg_parser.ts
Emulebest: Well, defaults are of type PapieaConfig, so they are expected to comply with config 100%

papiea-engine/src/utils/arg_parser.ts
Emulebest: I didn't get the solution exactly, would you like us to check that `PAPIEA_DEBUG` is either `"true"` or `"false"` and nothing else?

papiea-engine/src/utils/arg_parser.ts
Emulebest: Is it a valid TS code above? There are a few reasons I decided not to dive deep into the type system here:
1) We don't rely on TS types in runtime since we are using type erasure
2) The signature becomes needlessly complicated for any non inner circle maintainer

Emulebest: Found a simpler way of doing what you meant

papiea-engine/src/utils/arg_parser.ts
joshua-berry-ntnx: I think you missed what I was trying to say—right now, let's say I have defaults like `{foo: 1, bar: 2}`. Now let's say I have a config `{foo: 42}` and an environment variable `PAPIEA_BAR=17`. The way the code is written today, the `PAPIEA_BAR` variable (which is intended to override `bar`) will be ignored because `bar` is not specified in the config.

We should make sure environment variables can directly override defaults without needing to be specified in the config.

Emulebest: Okay, got it

PR #532, opened 2020-10-22
Limit error values in procedure invocation error

nitesh-idnani1:
Modified error fields initialization within the procedure invocation error type to limit redundant error fields


papiea-engine/src/errors/procedure_invocation_error.ts
```diff
@@ -7,10 +7,13 @@ export class ProcedureInvocationError extends Error {
     status: number;
 
     protected constructor(errors: { [key: string]: any }[], status: number) {
-        super(JSON.stringify(errors));
+        const messages = errors.map(x => x.message);
+        super(JSON.stringify(messages));
+        // super(errors[0].message);
```
Emulebest, line 12: Do we still need this comment?

Emulebest (changes requested):

PR #535, opened 2020-10-27
Randomness in intentful handler

### LINKED ISSUE #531: Dependant status handler doesn't work properly
### System Details

- **Papiea Engine version**: nutanix-docker.jfrog.io/papiea:0.7.10_1592
- **Papiea SDK version:** 1.0.90
- **Papiea Client version:** 1.0.90

### Describe the Bug
I have two fields on an entity: `netlinks` and `state`. Both have intent handlers assigned. State handler always waits for netlinks to be resolved prior it starts working. Both fields get updated at the same time (spec is updated) that causes 2 diff, one for `netlinks` and one for `state`. When netlinks intent handler is chosen to be resolved first, the system works as expected. When state intent handler is chosen to be resolved first, it runs over and over and never gets the netlinks intent handler to be resolved (as it never runs in such a case).

to put it generally, if we have a and b intent handler and b depends on a and they get updated at the same time, it doesn't work as expected when b is chosen first and works fine when a is chosen first. This happens because once a handler to be run is chosen it never lets the second one run between its runs.

### Bug information
FIelds schema:
```
        state: {
            type: 'string',
            enum: [
                // Machine is being provisioned
                State.Provisioning,
                // Machine is fully-provisioned and ready for use
                State.Online,
                // Machine has failed or terminated
                State.Terminated,
            ],
            description: 'Machine state',
        },
        netlinks: {
            type: 'array',
            description:
                'A list of network links configured for this instance',
            'x-papiea': 'status-only',
            items: {
                type: 'object',
                required: ['label', 'network_name', 'subnet_name'],
                properties: {
                    label: {
                        // TODO must be unique
                        type: 'string',
                        enum: [
                            // Required to assign AHV ip
                            NetlinkLabel.AHV,
                            // Required to assign CVM ip
                            NetlinkLabel.CVM,
                        ],
                        description:
                            `A user-defined string identifying a particular netlink`,
                    },
                    network_name: {
                        type: 'string',
                        description: 'Name of the network'
                    },
                    subnet_name: {
                        type: 'string',
                        description: 'Name of the subnet'
                    },
                    ip: {
                        type: 'string',
                        description: 'Allocated ip'
                    },
                },
            },
        },
```

intent handler declaration
```
    instance.on('state', ctxbuild.wrap_fn(Instance.on_state));
    instance.on('netlinks.+{label}',
        ctxbuild.wrap_fn(Instance.on_netlink_added));
```
Intent handlers definition (simplified)
```
export async function on_netlink_added(
    ctx: Context, instance: Entity, diff: Difference<Netlink>[]): Promise<void> {
    const subnet_kind = ctx.kind(Subnet.KIND)
    const instance_kind = ctx.kind(KIND)

    ctx.audit("diff", diff)
    for (const {spec: [spec]} of diff) {
        // TODO remove when http://github.com/nutanix/papiea/issues/527 is fixed
        const args: any = {workaround_dummy_string: ""}
        if (spec.ip){
            args.ip = spec.ip
        }
        const {ip} = await subnet_kind.invoke_procedure(
            'alloc_ip', spec.subnet, args);
        const new_netlink = {
            ...spec,
            ip: ip,
        }
        netlinks.push(new_netlink);
        await ctx.update_status(current_instance.metadata, {netlinks});
}
}

export async function go_online(
    ctx: Context, instance: Entity, state: State
): Promise<void> {
    if (!await can_go_online(ctx, instance)) {
        return
    }
}

async function can_go_online(ctx: Context, instance: Entity) {
    if (!deepEqual(instance.status.netlinks, instance.spec.netlinks)){
        ctx.info("The instance can't go online yet, " +
            "its netlinks have not been resolved in inter handler",
            {
                desired: instance.spec.netlinks,
                current: instance.status.netlinks
            })
        return false
    }
    return true
}
```

### Expected/Actual Behavior
We want either to be able to specify the order of intent handlers to be resolved or to make intent handlers run concurrently to each other. with two intent handlers a, b we expect the following timeline of invocation
`a b a b a b` and not `a a a a a a b b b b` or `b b b b b a a a a a`

### Steps to Reproduce

1. Add an entity with 2 fields
2. Add 2 intent handlers for them
3. Make one dependant on another resolution (check that the spec is the same as the state)
4. Observe that when the waiter handler runs first - things never resolve as (the dependant one never runs) and normal behavior when dependant runs first  

### Impact
For now I call the update 2 times in the order I want handlers to resolve, but it seems like a hack

<hr>

Emulebest:
resolves #531


shlomi-nx (commented):
Looks good to me!

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
```diff
@@ -71,7 +71,10 @@ describe("Intentful Workflow tests", () => {
         }
         first_provider_to_delete_entites = []
         second_provider_to_delete_entites = []
-        await providerApiAdmin.delete(`${first_provider_prefix}/${provider_version}`)
+        try {
+            await providerApiAdmin.delete(`${first_provider_prefix}/${provider_version}`)
+        } catch (e) {
+        }
```
shlomi-nx, line 77: Just silently ignoring may miss problems, no?

papiea-engine/__tests__/database_tests/database_mongo.test.ts
```diff
@@ -563,10 +569,10 @@ describe("MongoDb tests", () => {
                 kind: "test_kind"
             }
         }
-        watchlist.set([entry_ref, undefined, {delay: {delay_seconds: 120, delay_set_time: new Date()}, retries: 0}])
+        watchlist.set([entry_ref, [[{} as Diff, {delay: {delay_seconds: 120, delay_set_time: new Date()}, retries: 0}]]])
         await watchlistDb.update_watchlist(watchlist)
         const watchlistUpdated = await watchlistDb.get_watchlist()
-        expect(watchlistUpdated.get(entry_ref)![2]!.delay!.delay_seconds).toBe(120)
+        expect(watchlistUpdated.get(entry_ref)![1]![0]![1]!.delay!.delay_seconds).toBe(120)
```
shlomi-nx, line 575: Perhaps add comment on the mysterious `![1]![0]![1]`, i.e. why this is a correct assumption

papiea-engine/src/intentful_core/diff_selection_strategies/diff_selection_strategy_interface.ts
```diff
@@ -1,5 +1,7 @@
 import { Diff } from "papiea-core";
 
+type Index = number
+
 export interface DiffSelectionStrategyInterface {
-    selectOne: (diffs: Diff[]) => Diff
-}
\ No newline at end of file
+    selectOne: (diffs: Diff[]) => [Diff, Index]
```
shlomi-nx, line 6: Some comments on this function? I suppose it is returning the selected Diff as well as it's position?

papiea-engine/src/intentful_engine/diff_resolver.ts
```diff
@@ -226,28 +184,52 @@ export class DiffResolver {
...
+                diff_results[idx][1] = backoff
+            }
+        } else {
+            // Delay for rediffing
+            if ((new Date().getTime() - backoff.delay.delay_set_time.getTime()) / 1000 > backoff.delay.delay_seconds) {
+                const diff_index = rediff.diffs.findIndex(diff => deepEqual(diff.diff_fields, next_diff!.diff_fields))
+                // Diff still exists, we should check the health and retry if not healthy
+                if (diff_index !== -1) {
+                    try {
+                        // TODO: this seems like a bug
```
shlomi-nx, line 219: What does?

papiea-engine/src/intentful_engine/intent_resolver.ts
```diff
@@ -80,116 +84,75 @@ export class IntentResolver {
...
-                watcher.status = IntentfulStatus.Active
-                watcher.diffs = [...intersection]
-                await this.intentWatcherDb.update_watcher(watcher.uuid, { diffs: watcher.diffs })
+    private static determineWatcherStatus(affected_diff_count: number, resolved_diff_count: number, active: IntentWatcher): IntentfulStatus {
+        if (affected_diff_count > 0) {
+            if (resolved_diff_count > 0) {
+                return IntentfulStatus.Completed_Partially
+            }
+            if (affected_diff_count === active.diffs.length) {
+                return IntentfulStatus.Outdated
```
shlomi-nx, line 129: How is this logic correct?

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
```diff
@@ -1068,6 +1113,56 @@ describe("Intentful Workflow tests", () => {
             sdk.server.close();
         }
     })
+
+    test("Diff selection order should be random", async () => {
+        expect.assertions(2);
```
shlomi-nx, line 1118: How does this test verifies that the order is random?

shlomi-nx (commented):
Looks good, some comments:

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
Emulebest: This a cleanup code, pretty sure the deletion problems should be signified by earlier tests (entity.api.test.ts)

papiea-engine/__tests__/database_tests/database_mongo.test.ts
Emulebest: Yeah, I'll add the comment, though this is more or less obvious because we have an object declaration right above this piece of code

papiea-engine/src/intentful_core/diff_selection_strategies/diff_selection_strategy_interface.ts
Emulebest: Yes, the assumption is correct, will add a comment

papiea-engine/src/intentful_engine/diff_resolver.ts
Emulebest: Oh... right, forgot to check this, will check tomorrow. The healthcheck condition is incorrect logically as it is written right now

papiea-engine/src/intentful_engine/intent_resolver.ts
Emulebest: This means:
All the spec fields recorded by the watcher got changed as a series of spec changes but none of the diffs (on the watcher fields) got resolved, only affected, thus `Outdated`

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
Emulebest: Right, this naming is kind of a leftover. This test checks that multiple diffs could be tried at the same if one of the handlers fails and NOT get stuck in retrying single diff handler

papiea-engine/src/intentful_engine/diff_resolver.ts
Emulebest:
```
// TODO: this seems like a bug
if (await this.checkHealthy(rediff.diffs![diff_index])) {
     return
}
```
If diff handler is healthy it means that we could proceed retrying it, not return immediately

papiea-engine/src/intentful_engine/intent_resolver.ts
shlomi-nx: Got it, this will be a good comment to add to the code ;)

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
shlomi-nx: In case there's some problem with the cleanup, other tests may fail without an apparent reason. I think at least logging would be helpful.

papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts
shlomi-nx: Perhaps it's useful to add an explicit test, where a first diff is attempted to resolve, but can't, at which point another diff is entered into the system. Only after resolving the later diff the first one will get resolved. This will prove that the system will find a resolving path.

Also, please fix the names/titles of these tests to reflect what really is being tested.

PR #543, opened 2020-11-05
Add additional info in procedure invocation error

nitesh-idnani1:
Added stacktrace and errors field for validation error in procedure invocation and test to verify the same.

So far, I've added additional information for the `validation error` within the `Procedure Invocation Error` and will look at other places as well to ensure proper value for fields in error type.

### DIFF
--- a/papiea-engine/src/errors/procedure_invocation_error.ts
+++ b/papiea-engine/src/errors/procedure_invocation_error.ts
@@ -25,9 +25,13 @@ export class ProcedureInvocationError extends Error {
                 stacktrace: err.response?.data.stacktrace
             }], err.response?.status ?? 500)
         } else if (err instanceof ValidationError) {
-            return new ProcedureInvocationError(err.errors.map(e => {
-                return { message: e }
-            }), status || 500)
+            return new ProcedureInvocationError(
+                err.errors.map(e => ({
+                    message: e,
+                    errors: {},
+                    stacktrace: err.stack
+                }))
+            , status || 500)
         } else {
             return new ProcedureInvocationError([{
                 message: "Unknown error during procedure invocation",
--- a/papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
+++ b/papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
@@ -743,6 +743,8 @@ describe("Provider Sdk tests", () => {
             const res: any = await axios.post(`${sdk.entity_url}/${sdk.provider.prefix}/${sdk.provider.version}/procedure/computeSum`, { input: { "a": 5, "b": 5 } });
         } catch (e) {
             expect(e.response.data.error.errors[0].message).toBe("Unable to validate a model with a type: string, expected: number");
+            expect(e.response.data.error.errors[0].stacktrace).not.toBeUndefined();
+            expect(e.response.data.error.errors[0].stacktrace).toContain("Unable to validate a model with a type: string, expected: number")
             expect(e.response.data.error.code).toBe(500);
         } finally {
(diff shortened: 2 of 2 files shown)


papiea-engine/src/errors/procedure_invocation_error.ts
```diff
@@ -25,9 +25,15 @@ export class ProcedureInvocationError extends Error {
...
-                return { message: e }
-            }), status || 500)
+            return new ProcedureInvocationError(
+                err.errors.map(e => {
+                    return {
+                        message: e,
+                        errors: {},
+                        stacktrace: err.stack
+                    }
+                })
```
joshua-berry-ntnx, line 35: One cosmetic note: Rather than writing an explicit return, you can just wrap the returned object in parentheses, which is a little simpler/more readable:

```ts
err.errors.map(e => ({message: e, errors: {}, ...}))
```

shlomi-nx:
@nitesh-idnani1 In the description of a PR, you should always contain something like `fixes {issue number}`. This will both allow the reader to quickly get the context of the PR, as well as allow github to automatically close the issue when the PR is merged.

shlomi-nx (commented):
Perhaps the SDK should append to the stacktrace so that it is easy for users to see where in their code the invocation happened.

nitesh-idnani1:
> @nitesh-idnani1 In the description of a PR, you should always contain something like `fixes {issue number}`. This will both allow the reader to quickly get the context of the PR, as well as allow github to automatically close the issue when the PR is merged.

I've added the issue in the linked issues on this page which has done the required things of mentioning PR in that issue and closing it when this PR gets merged. I can add in the description as well if it helps

papiea-engine/src/errors/procedure_invocation_error.ts
nitesh-idnani1: Fixed this, thanks

shlomi-nx:
Ohh, I suppose I am outdated by now :) It used to just be that `fixed {number}` description.. It's ok, thanks for pointing me out to this!

nitesh-idnani1:
> Perhaps the SDK should append to the stacktrace so that it is easy for users to see where in their code the invocation happened.

In this request, since we are using axios there is no stack trace coming from the sdk end. but I will look at the sdk and add trace where it's missing them.

papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -743,6 +743,8 @@ describe("Provider Sdk tests", () => {
             const res: any = await axios.post(`${sdk.entity_url}/${sdk.provider.prefix}/${sdk.provider.version}/procedure/computeSum`, { input: { "a": 5, "b": 5 } });
         } catch (e) {
             expect(e.response.data.error.errors[0].message).toBe("Unable to validate a model with a type: string, expected: number");
+            expect(e.response.data.error.errors[0].stacktrace).not.toBeUndefined();
+            expect(e.response.data.error.errors[0].stacktrace).toContain("Unable to validate a model with a type: string, expected: number")
```
shlomi-nx, line 747: Perhaps not for this PR, but this message is way to generic. It does not point you to which item in your input has the problem, users don't know what "model" stands for.. Let's create a ticket to improve validation error reporting.

shlomi-nx (approved):

papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
nitesh-idnani1: Sure, we should do it.

PR #546, opened 2020-11-12
Added module to remove status-only fields from entity status

nitesh-idnani1:
Added module to loop through the entity schema and remove all the `status-only` fields from the entity status object.
Added test to verify the functionality

### DIFF
--- a/papiea-engine/src/intentful_engine/diff_resolver.ts
+++ b/papiea-engine/src/intentful_engine/diff_resolver.ts
@@ -99,49 +99,22 @@ export class DiffResolver {
     private async rediff(entry_reference: EntryReference): Promise<RediffResult | null> {
         try {
             const [metadata, spec] = await this.specDb.get_spec(entry_reference.entity_reference)
-            let [, status] = await this.statusDb.get_status({...entry_reference.provider_reference, ...entry_reference.entity_reference})
+            const [, status] = await this.statusDb.get_status({...entry_reference.provider_reference, ...entry_reference.entity_reference})
             const provider = await this.providerDb.get_provider(entry_reference.provider_reference.provider_prefix, entry_reference.provider_reference.provider_version)
             const kind = this.providerDb.find_kind(provider, metadata.kind)
-            const schemas: any = Object.assign({}, kind.kind_structure);
-            status = this.remove_status_only_fields(schemas[kind.name], status);
+            let diff_status: any = {}
+            if (status !== undefined) {
+                diff_status = JSON.parse(JSON.stringify(status));
+            }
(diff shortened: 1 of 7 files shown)


papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -1058,6 +1059,76 @@ describe("Provider Sdk tests", () => {
...
+                    items: {
+                        type: 'object',
+                        properties: {
+                            id: {
+                                type: 'string'
+                            },
+                            a: {
+                                'x-papiea': 'status-only',
+                                type: 'string'
+                            }
```
shlomi-nx, line 1081: I'd add another field that is not `status-only`, otherwise you are hitting a base case where outside of the identifier the content is empty. By having `b` which is the same and `a` which is different but status only we can make sure logic works better.

shlomi-nx (commented):
Looks good, please see comment about test below.

papiea-engine/src/intentful_engine/diff_resolver.ts
```diff
@@ -112,6 +114,36 @@ export class DiffResolver {
         }
     }
 
+    private remove_status_only_fields(schema: any, status: Status): Status {
```
Emulebest, line 117: I am not sure this fits here (in diff resolver), I think this kind of cleanup should be done in the `Differ` specific implementation. Some day we might want to have a differ which is aware of schema

Emulebest (changes requested):

papiea-engine/src/intentful_engine/diff_resolver.ts
```diff
@@ -102,8 +102,11 @@ export class DiffResolver {
             const [, status] = await this.statusDb.get_status({...entry_reference.provider_reference, ...entry_reference.entity_reference})
             const provider = await this.providerDb.get_provider(entry_reference.provider_reference.provider_prefix, entry_reference.provider_reference.provider_version)
             const kind = this.providerDb.find_kind(provider, metadata.kind)
+            const schemas: any = Object.assign({}, kind.kind_structure);
+            let diff_status = Object.assign({}, status);
```
Emulebest, line 106: Damn Javascript, one more quirk here. Object.assign() only clones a single level of object depth, so I suggest using `JSON.parse(JSON.stringify(status))` here to clone the whole object

Emulebest (changes requested):
A single remark, other than that good job!

nitesh-idnani1:
Hey @Emulebest
Can you review this PR and merge it, thanks.

Emulebest (approved):

PR #548, opened 2020-11-16
Fix swagger errors in papiea-engine

nitesh-idnani1:
Fix errors related to the admin swagger and other errors within the api-docs code related to swagger

### DIFF
--- a/papiea-engine/__tests__/docs_tests/api_docs.test.ts
+++ b/papiea-engine/__tests__/docs_tests/api_docs.test.ts
@@ -1,5 +1,5 @@
 import "jest"
-import { unlinkSync, writeFileSync } from "fs";
+import { readFileSync, unlinkSync, writeFileSync } from "fs";
 import { validate } from "swagger-parser";
 import axios from "axios"
 import {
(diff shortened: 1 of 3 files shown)


shlomi-nx:
A little difficult understanding the diff, can you point out the problems you have found and fixed?

nitesh-idnani1:
> A little difficult understanding the diff, can you point out the problems you have found and fixed?

Yes, the post request was giving the following error due to the missing `responses` field in the schema,

```
Structural error at paths./services/test_provider/0.1.0/Test.post
should have required property 'responses'
missingProperty: responses
```

shlomi-nx:
Got it, thanks! If you now make some provider you are seeing no errors at all?

nitesh-idnani1:
> Got it, thanks! If you now make some provider you are seeing no errors at all?
> 
> Edit:
> Perhaps we should add some test that verifies that there are no swagger schema errors?

I haven't tried with a new custom schema, could you maybe send me the bmaas one privately so that I can verify if we're missing something in papiea?

Yes, we have a few tests [here](https://github.com/nutanix/papiea/blob/master/papiea-engine/__tests__/docs_tests/api_docs.test.ts) and I'm working on adding more tests to validate the schema.

shlomi-nx:
Sorry, I can't send you bmaas's one. But it should not depend on it, just make up some provider definition to test this, just make sure it has representatives of all elements (kinds, spec/status-only, procedures of various levels). Nothing really needs to be implemented, just registered so that we get a schema.

shlomi-nx (approved):
Looks good

PR #549, opened 2020-11-16
Changed error message for bad request error in validator

nitesh-idnani1:
- Modified the error message thrown by validator due to missing body/query params in procedure request.

### DIFF
--- a/papiea-engine/src/validator/express_validator.ts
+++ b/papiea-engine/src/validator/express_validator.ts
@@ -12,13 +12,13 @@ export function check_request(request_validator_options: RequestValidatorOptions
     return (req: Request, res: Response, next: NextFunction) => {
         for (var req_query_param in req.query) {
             if (!allowed_query_params_set.has(req_query_param)) {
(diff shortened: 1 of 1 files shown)


papiea-engine/src/validator/express_validator.ts
```diff
@@ -12,13 +12,13 @@ export function check_request(request_validator_options: RequestValidatorOptions
     return (req: Request, res: Response, next: NextFunction) => {
         for (var req_query_param in req.query) {
             if (!allowed_query_params_set.has(req_query_param)) {
-                throw new BadRequestError(`Allowed query params ${request_validator_options.allowed_query_params}`);
+                throw new BadRequestError(`Request is missing query parameter: ${request_validator_options.allowed_query_params}`);
```
Emulebest, line 15: The code doesn't imply the error message. I would say a better error could be something like:
`Parameter ${param} is not allowed. Allowed params: ${allowed_params}`

Emulebest (changes requested):

PR #550, opened 2020-11-23
Custom entity constructors implementation

### LINKED ISSUE #505: Custom entity constructors
Current `on-create` function is invoked on a newly-created entity by providing all spec fields. There are a number of reasons why this is not sufficient, among others we have the need to establish some constants which will be reflected as `status-only` fields, or the need to provide a higher level of abstraction than individual and exhaustive spec definition.

The idea is to leverage the existing procedures mechanism to establish a custom constructor.

# Definition

a custom constructor function is a function of type: `ConstructorCtx, Input->Entity`:
1. `ConstructorCtx` is a similar a procedure's context, with some additions, expanded below
2. `Input` is a custom schema defined in the same way as any procedure does
3. `Entity` is a 3-tuple consisting of the following:
  a. Partial `metadata` section - in the same way current entity creation POST accepts
  b. `spec` section - in the same way current entity create POST accepts
  c. `status` section - all the status fields to be set on this new entity

The custom constructor function will be invoked prior to the actual entity creation, and the returned `Entity` object will be created in Papiea.

`ConstructorCtx` will provide the ability to register a callback function that will execute after successful creation of the entity. 

# Flow
1. If the custom constructor function fail to return a valid `Entity` representation, no entity gets created and a proper error message and code is returned.
2. If the optional post-create function throws an error, the just created entity will get deleted and a proper error code will be returned. the failed post-create function should handle all cleanup as well
3. If both functions succeeds, the created `Entity` is returned to the user
4. Only after both functions succeeds, the entity is available for others to use (it is available for GET, PUT, DELETE)

# Example
```typescript

const the_kind = sdk.kind(...)

the_kind.on_create(InputSchema, (ctx: ConstructorCtx, input: Input):Entity => {
    // 1. Define the entity, run code etc
    let entity: Entity= {
        metadata: {uuid: "...", extension: {owner: "me"}},
        spec: {...},
        status: {...}
    }

    // 2. Register post creation closure, which wraps around the current variables
    ctx.post_create((created_entity: Entity):None->{
        // 4. Post create is a callback after creation:

        ctx.update_status(created_entity.metadata, {...})
    })

    // 3. Now create the entity
    return entity
}
```

# Exported swagger schema for the constructor
The default entity's create POST will be replaced by a the POST procedure registered using the `on-create` function. It's input will be that of the `Input` schema, and it's output will be a full `Entity`, the same as returned by `GET`.

No other `procedures` will be generated for the custom constructor, e.g. no `on-create` or `post-create` entries should be exposed via swagger spec.

If no custom constructor is defined, standard POST constructor will be created, as done today.

Emulebest:
resolves #505


papiea-core/src/core.ts
```diff
@@ -70,8 +70,6 @@ export interface Provider_Entity_Reference extends Entity_Reference {
 // [[file:~/work/papiea-js/Papiea-design.org::#h-Metadata-350][metadata-struct]]
 export interface Metadata extends Provider_Entity_Reference {
     // Identity fields
-    uuid: uuid4;
-    kind: string;
```
shlomi-nx, line 74: Why did you remove uuid and kind from metadata? I suppose it exists in the reference

papiea-engine/__tests__/auth_tests/entity_api_auth.test.ts
```diff
@@ -211,17 +222,7 @@ describe("Entity API auth tests", () => {
...
-    });
-
-    test("Get user info", async () => {
-        expect.hasAssertions();
-        const { data: { token } } = await providerApi.get(`/${ provider.prefix }/${ provider.version }/auth/login`);
-        const { data } = await providerApi.get(`/${ provider.prefix }/${ provider.version }/auth/user_info`,
-            { headers: { 'Authorization': 'Bearer ' + token } }
-        );
-        expect(data.owner).toEqual("alice");
-        expect(data.tenant).toEqual(tenant_uuid);
```
shlomi-nx, line 224: Why were tests removed?

papiea-engine/src/api_docs/api_docs_generator.ts
```diff
@@ -243,7 +277,7 @@ export default class ApiDocsGenerator {
 
     postKind(provider: Provider, kind: Kind) {
         return {
-            "description": `Creates a new ${ kind.name }`,
+            "description": `Creates a new ${ kind.name }. While constructing copies spec to status`,
```
shlomi-nx, line 280: Is this text for the default constructor? It should say something like:
`description: Creates a new instance of {kind}. The created entity will have no diffs upon creation since this default constructor sets status to be the same as spec`

papiea-engine/src/api_docs/api_docs_generator.ts
```diff
@@ -460,6 +494,33 @@ export default class ApiDocsGenerator {
         return this.getProcedureSchema(base, procedure)
     }
 
+    callConstructorProcedure(provider: Provider, kind: Kind, procedure: Procedural_Signature) {
+        const [input, _] = this.getKindProcedureSchema(provider, kind, procedure)
+        const procedural_def = {
+            "description": `Creates a new ${ kind.name }`,
```
shlomi-nx, line 500: `description: Creates a new instance of {kind} based on the provided input parameters. The created entity may have diffs upon creation which will get resolved in the usual way.`

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -61,85 +58,81 @@ export class Entity_API_Impl implements Entity_API {
 
     async get_intent_watcher(user: UserAuthInfo, id: string): Promise<Partial<IntentWatcher>> {
         const intent_watcher = await this.intent_watcher_db.get_watcher(id)
-        const [metadata, _] = await this.spec_db.get_spec(intent_watcher.entity_ref)
-        await this.authorizer.checkPermission(user, { "metadata": metadata }, Action.Update);
+        // TODO: this should be a separate authorizer not a PerProvider one
+        // await this.authorizer.checkPermission(user, { "metadata": metadata }, Action.Update);
```
shlomi-nx, line 62: Please explain. I see some changes in authorization in this PR, can you explain a little?

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -61,85 +58,81 @@ export class Entity_API_Impl implements Entity_API {
...
     }
 
-    async filter_entity_spec(user: UserAuthInfo, kind_name: string, fields: any, exact_match: boolean, sortParams?: SortParams): Promise<[Metadata, Spec][]> {
+    async filter_entity_spec(user: UserAuthInfo, prefix: string, version: Version, kind_name: string, fields: any, exact_match: boolean, sortParams?: SortParams): Promise<[Metadata, Spec][]> {
+        const provider = await this.providerDb.get_provider(prefix, version);
         fields.metadata.kind = kind_name;
         const res = await this.spec_db.list_specs(fields, exact_match, sortParams);
-        const filteredRes = await this.authorizer.filter(user, res, Action.Read, x => { return { "metadata": x[0] } });
+        const filteredRes = await this.authorizer.filter(user, res, Action.Read, provider, x => {
+            return {"metadata": x[0]}
```
shlomi-nx, line 114: is this sufficient? won't it just return `{'metadata': {..}}`? What about the specs (and status in the next function)?

shlomi-nx (commented):
Went briefly over these commits. Looks good, I left a few questions.

papiea-engine/__tests__/auth_tests/entity_api_auth.test.ts
Emulebest: Didn't see that, probably a bug while uncommenting in my IDE

papiea-engine/src/api_docs/api_docs_generator.ts
Emulebest: Okay, will change the message

papiea-engine/src/api_docs/api_docs_generator.ts
Emulebest: Okay, will change that

papiea-engine/src/entity/entity_api_impl.ts
Emulebest: PerProvider authorizer is used to check items with metadata extensions to belong to particular provider. Watcher are firstly global (not in provider scope), thus should use a different authorizer

papiea-engine/src/entity/entity_api_impl.ts
Emulebest: It is a custom filter function so it functions as intended :) Please see https://github.com/nutanix/papiea/blob/efeb6927d0a9d3e1881b2b62963c6e834660db59/papiea-engine/src/entity/entity_routes.ts#L30

papiea-core/src/core.ts
Emulebest: Yes, it is already inherited from the reference structure

papiea-engine/src/entity/entity_api_impl.ts
shlomi-nx: Do you plan to introduce some other authorizer for that? It should essentially be nothing but the papiea-admin s2skey, right?

Emulebest: Nope since we might give different rights to different users to observe different watchers. (e.g. The user who created the diff gets to view it)

shlomi-nx: hmm I see. Thanks for the details.

joshua-berry-ntnx (commented):
General comments; I'll leave it to the actual Papiea devs to approve or request changes.

papiea-engine/src/intentful_core/entity_creation_strategies/constructor_entity_creation_strategy.ts
```diff
@@ -0,0 +1,120 @@
...
+                await this.watchlistDb.update_watchlist(watchlist)
+            }
+        }
+        return [watcher, [created_metadata, created_spec, created_status]]
+    }
+
+    protected async invoke_constructor(procedure_name: string, input: any): Promise<Entity> {
+        let entity: Entity
+        const constructor = this.kind.kind_procedures[procedure_name]
+        if (this.kind) {
```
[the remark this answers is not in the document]
shlomi-nx: I would read @joshua-berry-ntnx's comment as a request to move line 82 after line 83.

papiea-engine/src/intentful_core/entity_creation_strategies/entity_creation_strategy_interface.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    protected async create_entity(metadata: Metadata, spec: Spec): Promise<[Metadata, Spec]> {
+        // Create increments spec version so we should check already incremented one
+        await this.check_spec_version(metadata, metadata.spec_version + 1, spec)
+        const [updatedMetadata, updatedSpec] = await this.specDb.update_spec(metadata, spec);
+        await this.statusDb.replace_status(metadata, spec)
+        return [updatedMetadata, updatedSpec]
+    }
+
+    abstract create(input: unknown): Promise<[IntentWatcher | null, [Metadata, Spec, Status | null]]>
```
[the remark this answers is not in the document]
shlomi-nx: I agree, and more, it becomes a pain to make any changes.

papiea-engine/src/intentful_core/intentful_context.ts
```diff
@@ -67,4 +77,17 @@ export class IntentfulContext {
...
         return strategy
     }
+
+    getEntityCreationStrategy(provider: Provider, kind: Kind, user: UserAuthInfo): EntityCreationStrategy {
+        let strategy: EntityCreationStrategy | undefined
+        if (kind.kind_procedures[`__${ kind.name }_create`] !== null && kind.kind_procedures[`__${ kind.name }_create`] !== undefined) {
+            strategy = this.entityCreationStrategyMap.get("constructor")
+        } else {
+            strategy = this.entityCreationStrategyMap.get("basic")
+        }
```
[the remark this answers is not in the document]
Emulebest: Extensibility purposes

papiea-engine/src/provider/provider_api_impl.ts
```diff
@@ -133,6 +133,20 @@ export class Provider_API_Impl implements Provider_API {
...
+                status = {}
+            } else if (schemas.type === 'object' && schemas.properties) {
+                const properties_schema = schemas.properties;
+                for (var key in status) {
+                    status[key] = this.fix_null_object(properties_schema[key], status[key])
+                }
+            }
+        }
+        return status
+    }
```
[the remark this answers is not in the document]
Emulebest: I think one of the merges added it here, I'll remove this

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -571,17 +571,18 @@ export class Kind_Builder {
         return this
     }
 
-    on_create(handler: (ctx: ProceduralCtx_Interface, entity: Partial<Entity>) => Promise<any>): Kind_Builder {
+    // Return type should always contain metadata & spec (status could be empty)
+    on_create(description: {input_schema: any, error_schemas?: ErrorSchemas}, handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<{spec: Spec, status: Status, metadata?: Partial<Metadata>}>): Kind_Builder {
         const name = `__${this.kind.name}_create`
         const loggerFactory = new LoggerFactory({logPath: name})
         const [logger, handle] = loggerFactory.createLogger()
         logger.info("You are registering on create handler. Note, this is a post create handler. The behaviour is due to change")
```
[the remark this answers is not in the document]
Emulebest: Thanks, good catch :)

papiea-engine/src/intentful_core/intentful_context.ts
```diff
@@ -67,4 +77,17 @@ export class IntentfulContext {
...
+        strategy?.setKind(kind)
+        strategy?.setUser(user)
+        strategy?.setProvider(provider)
```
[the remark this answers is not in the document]
Emulebest: Well, this comes down to the single threaded nature of node js. We actually put an implicit suspension point in front of every async call that uses the strategy:
`const [intent_watcher, [metadata, spec, status]] = await strategy.create(input)`
So we are not going to be having a shared access in this case. However if it'd would look like this:
`const [intent_watcher, [metadata, spec, status]] = strategy.create(input)`
Then we could potentially context switch before the first internal suspension point and get a bad case of shared resource inconsistency

papiea-engine/src/intentful_core/entity_creation_strategies/entity_creation_strategy_interface.ts
```diff
@@ -0,0 +1,130 @@
...
+            const result_spec = await this.specDb.list_specs({ metadata: { uuid: uuid, kind: kind_name, provider_version: provider.version, provider_prefix: provider.prefix, deleted_at: null } }, false)
+            const result_status = await this.statusDb.list_status({ metadata: { uuid: uuid, kind: kind_name, provider_version: provider.version, provider_prefix: provider.prefix, deleted_at: null } }, false)
+            if (result_spec.length !== 0 || result_status.length !== 0) {
+                return [result_spec[0][0], result_spec[0][1], result_status[0][1]]
+            } else {
+                return []
+            }
+        } catch (e) {
+            // Hiding details of the error for security reasons
+            // since it is not supposed to occur under normal circumstances
```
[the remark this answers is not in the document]
Emulebest: Good point, I'll do this in a separate logging PR that should come fairly soon

papiea-engine/src/intentful_core/entity_creation_strategies/basic_entity_creation_strategy.ts
```diff
@@ -0,0 +1,30 @@
...
+        await this.validate_entity({metadata, spec: input.spec, status: input.spec})
+        const [created_metadata, spec] = await this.create_entity(metadata, input.spec)
+        if (this.kind?.intentful_behaviour === IntentfulBehaviour.Differ) {
+            const watchlist = await this.watchlistDb.get_watchlist()
+            const ent = create_entry(created_metadata)
+            if (!watchlist.has(ent)) {
+                watchlist.set([ent, []])
+                await this.watchlistDb.update_watchlist(watchlist)
+            }
+        }
```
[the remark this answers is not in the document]
Emulebest: They are a bit different and are going to get even more different with the enforcement of `spec-only` rules on the constructor strategy

papiea-engine/src/intentful_core/entity_creation_strategies/entity_creation_strategy_interface.ts
```diff
@@ -0,0 +1,130 @@
...
+export abstract class EntityCreationStrategy {
+    protected readonly specDb: Spec_DB
+    protected readonly statusDb: Status_DB
+    protected readonly graveyardDb: Graveyard_DB
+    protected readonly watchlistDb: Watchlist_DB
+    protected readonly validator: Validator
+    protected readonly authorizer: Authorizer
+    protected kind!: Kind
+    protected user!: UserAuthInfo
+    protected provider!: Provider
```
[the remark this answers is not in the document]
Emulebest: The semantics of this property syntax is different. This is used to tell the compiler that I am certain that all of those properties are going to be set by the time I dereference them

papiea-sdk/python/e2e_tests/procedure_handlers.py
```diff
@@ -48,9 +50,8 @@ async def ensure_bucket_exists(ctx, input_bucket_name):
                 kind=ret_entity.metadata.kind
             )
     except:
-        raise
+        raise Exception("Unable to create bucket entity")
```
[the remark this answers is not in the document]
Emulebest: I think this should go with the logging PR, this one is already too big IMO

PR #553, opened 2020-11-30
Removed 'required' tag from schema passed to swagger validator

nitesh-idnani1:
- Remove 'required' tag from the schema passed to the swagger validate function inside validate_status module.
- Added test to verify the same i.e. partial status update with a missing required field.

### DIFF
--- a/papiea-engine/src/validator/index.ts
+++ b/papiea-engine/src/validator/index.ts
@@ -107,7 +107,8 @@ export class ValidatorImpl {
         if (kind === undefined) {
             throw new Error("Kind not found");
         }
-        const schemas: any = Object.assign({}, kind.kind_structure);
+        let schemas: any = Object.assign({}, kind.kind_structure);
+        schemas[kind.name] = remove_required_tag(schemas[kind.name])
         this.validate(status, Object.values(kind.kind_structure)[0], schemas,
             allowExtraProps, Object.keys(kind.kind_structure)[0]);
     }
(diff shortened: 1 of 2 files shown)


shlomi-nx (commented):
Is this only removing the required from the status portion? Validator should definitely check for required `spec` fields.

nitesh-idnani1:
> Is this only removing the required from the status portion? Validator should definitely check for required `spec` fields.

Yes, this just removes the required tag from schema for the status validate function and nowhere else.

Emulebest (commented):
@shlomi-nx the PR is good, but I don't think this is a correct solution. We essentially silently removing one of validation properties (requiredness). This is not only formally unsound but also could introduce misconceptions for provider writers.
I'd suggest:
a) Making this removal a flag while creating a provider
b) Splitting definition of spec and status on demand

shlomi-nx:
@Emulebest While I agree with your point in the general case, in this case we are trying to solve a problem where a spec-update request by the user may get rejected due to missing status fields - something the user have no control of and thus should not fail for him. 

The original idea was only for spec to be `requireable` but further discussions with @joshua-berry-ntnx made us keep them as well. In any event, there is no point at all checking status when receiving a spec change. `status` should be checked when provider update's status though, but it might be difficult to do with PATCHs.

Emulebest:
@shlomi-nx @nitesh-idnani1 What are the next steps for this one since constructors are getting very close?

nitesh-idnani1:
> @shlomi-nx @nitesh-idnani1 What are the next steps for this one since constructors are getting very close?

We can discuss this next time.

PR #555, opened 2020-12-03
Separated null value fields and unset in status db

nitesh-idnani1:
- Separated null valued fields and unset them in status db update.
- Removed fix null objects module which replaced null value with `{}`.


shlomi-nx (commented):
Looks good!

Emulebest (approved):

PR #564, opened 2020-12-09
Added deprecate info for the replace_status sdk method and relavant tests

nitesh-idnani1:
- Added info about the function and deprecate tag for the replace_status sdk method.
- Added tests to verify the update_status method and ensure similarity to replace_status.
- Modified mongo update_status method to use upsert flag for update call.

### DIFF
--- a/papiea-sdk/typescript/src/provider_sdk/typescript_sdk_context_impl.ts
+++ b/papiea-sdk/typescript/src/provider_sdk/typescript_sdk_context_impl.ts
@@ -72,19 +72,30 @@ export class ProceduralCtx implements ProceduralCtx_Interface {
             status: status
         });
         if (res.status != 200) {
-            console.error("Could not update status:", entity_reference, status, res.status, res.data);
+            this.get_logger().error("Could not update status:", entity_reference, status, res.status, res.data);
             return false
         }
         return true
     }
 
+    /**
+     * Function which calls post on the update_status endpoint
+     * which internally calls the mongo update function with upsert
+     * flag set to true.
+     *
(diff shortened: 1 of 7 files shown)


nitesh-idnani1:
I have not specified the version after which replace_status would be deleted (for now I've kept it as 'y'), so we need to make that decision.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk_context_impl.ts
```diff
@@ -78,7 +78,18 @@ export class ProceduralCtx implements ProceduralCtx_Interface {
...
+     * which internally calls the mongo update function with upsert
+     * flag set to true.
+     *
+     * This functions inserts a new document if no matching document
+     * is found for the query.
+     *
+     * @deprecated Will be deleted in version x. Use update_status instead.
+    */
     async replace_status(entity_reference: Entity_Reference, status: Status, provider_prefix: string = this.provider.get_prefix(), provider_version: Version = this.provider.get_version()): Promise<boolean> {
+        console.warn("Calling a deprecated method!!!")
```
Emulebest, line 92: Please use a logger instead

Emulebest (commented):
@shlomi-nx @nitesh-idnani1  Since we now using version `0.8.x` and constructors will be `0.9.x`, I propose something around `0.10.0` or `0.11.0` as version where we plan to delete this

shlomi-nx (commented):
Looks good

Emulebest:
@nitesh-idnani1 Please resolve the conflicts and put 0.11.0 as a deprecation version number

PR #569, opened 2020-12-11
Added validation check to not allow required status-only fields in schema

nitesh-idnani1:
- Added validation function to throw error if a required `status-only` field exists in schema.
- Added validation function to throw error is create/update spec contains value for a `status-only` field.
- Added tests to verify the above validation functions.

### DIFF
--- a/papiea-engine/src/validator/index.ts
+++ b/papiea-engine/src/validator/index.ts
@@ -94,8 +94,24 @@ export class ValidatorImpl {
 
     public validate_spec(spec: Spec, kind: Kind, allowExtraProps: boolean) {
         const schemas: any = Object.assign({}, kind.kind_structure);
+        // remove any status-only field from the schema to pass to validator
+        this.remove_schema_fields(schemas, "status-only")
         this.validate(spec, Object.values(kind.kind_structure)[0], schemas,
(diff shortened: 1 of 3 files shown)


papiea-engine/src/validator/index.ts
```diff
@@ -202,6 +219,42 @@ export class ValidatorImpl {
...
+                    }
+                    this.validate_status_only_field(field_schema["properties"])
+                }
+                if (field_schema.type === "array") {
+                    if (field_schema["items"]["type"].includes("object") && field_schema["items"].hasOwnProperty("required") && field_schema["items"].hasOwnProperty("properties")) {
+                        for (let req_field of field_schema["items"]["required"]) {
+                            if (field_schema["items"]["properties"][req_field].hasOwnProperty("x-papiea") && field_schema["items"]["properties"][req_field]["x-papiea"] === "status-only") {
+                                throw new ValidationError([{
+                                    name: "ValidationError",
+                                    message: `${req_field} of type status-only is set to be required`
```
Emulebest, line 245: Maybe something like `${req_field} is status-only but set to be required. Only spec fields could be required`

Emulebest (commented):
Seems good, please take a look at the comment

shlomi-nx (approved):
LGTM

PR #570, opened 2020-12-14
Fixed payload on client to not collide with constructor implementation details

Emulebest:


papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -1776,7 +1776,7 @@ describe("SDK callback tests", () => {
         }
     });
 
-    test("Client should infer request format when on create callback is NOT present", async () => {
+    test("Client should use appropriate request format when on create callback is NOT present", async () => {
```
joshua-berry-ntnx, line 1779: Is this really testing anything different to the callback-present case?

joshua-berry-ntnx (approved):

papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
Emulebest: It still tests that both regular and constructor creates succeed in creating the entity

joshua-berry-ntnx: Right, but isn't that more of an engine test than an SDK test? (I'm fine keeping it, just trying to see if we are duplicating anything)

PR #592, opened 2021-01-13
Removed requirement for input field in the procedures input.

nitesh-idnani1:
- Removed requirement for input field in procedure's input.
- Modified and fixed tests to use the new input schema changes for procedure calls.

### DIFF
--- a/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
+++ b/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
@@ -121,7 +121,7 @@ describe("Procedures tests", () => {
             const res = err.response;
             expect(res.status).toEqual(400);
             expect(res.data.error.errors.length).toEqual(1);
-            expect(res.data.error.errors[0].message).toBe("Unable to validate an empty value for property: rootModel")
+            expect(res.data.error.errors[0].message).toBe("moveX with schema MoveInput was expecting non-empty input");
         }
     });
     test("Procedure result validation", async () => {
--- a/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
+++ b/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
@@ -61,7 +61,7 @@ describe("Procedures tests", () => {
                 });
                 req.on('end', function () {
                     const post = JSON.parse(body);
-                    post.spec.x += post.input.x;
+                    post.spec.x += parseInt(post.input);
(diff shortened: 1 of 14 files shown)


Emulebest:
Is this one ready? If not please link why and maybe convert to draft

nitesh-idnani1:
> Is this one ready? If not please link why and maybe convert to draft

Yes, converting this to in progress and will be completed once we finalize the idea mentioned in the input schema discussion.
#593

papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
```diff
@@ -121,7 +121,7 @@ describe("Procedures tests", () => {
             const res = err.response;
             expect(res.status).toEqual(400);
             expect(res.data.error.errors.length).toEqual(1);
-            expect(res.data.error.errors[0].message.includes("undefined")).toBeTruthy();
+            expect(res.data.error.errors[0].message).toBe("Unable to validate an empty value for property: rootModel")
```
Emulebest, line 124: This error message signifies that something has gone off the roads here :). This is one of our most popular bugs `Unable to validate an empty value for property: rootModel` and should be addressed for this particular case (empty input)

Emulebest (changes requested):

papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
nitesh-idnani1: So, we should be throwing a specific error message for the empty input case rather than the above one, is that what you mean?

Emulebest: Correct

nitesh-idnani1: Okay, I'll look into it. Thanks

nitesh-idnani1: Added check in the validate function to throw error if model is not nullable (i.e. has required fields) and the input is empty/null. Kindly check and let me know if we need anything else. Thanks

nitesh-idnani1:
Reverted the new commit since I and @joshua-berry-ntnx has a discussion and decided it would be a bad practice to specify non-json or non-object like input to the procedures.

@shlomi-nx
Feel free to add your thoughts to this. Thanks

nitesh-idnani1:
This PR has the changes discussed today. Kindly review and merge it. Thanks

Emulebest:
@nitesh-idnani1 Regarding the recent bugs, could you also merge master into this branch and manually verify that api-docs are correctly rendered

PR #594, opened 2021-01-19
Reduced diff retry backoff exponent for debugging purposes

nitesh-idnani1:
- Added papiea debug variable in diff resolver to use in diff retry logic.
- Reduced retry backoff exponent to `1.3` if debug flag is set in engine.
- Modified tests to adjust to the above changes.


joshua-berry-ntnx:
I don't think we should do this just in debug builds, I think it's quite relevant for release builds as well.

nitesh-idnani1:
> I don't think we should do this just in debug builds, I think it's quite relevant for release builds as well.

Yes, I am thinking about two things i.e.
1. What would be an ideal value for this considering the latency and workflows for papiea, maybe @Emulebest can provide more inputs on this?
2. If (and how) we should allow the user to customize the value for the backoff exponent?

papiea-engine/src/intentful_engine/diff_resolver.ts
```diff
@@ -74,7 +74,7 @@ export class DiffResolver {
         }
     }
 
-    private createDiffBackoff(kind: Kind, delay: Delay | null | undefined, retries: number = 0): Backoff {
+    private createDiffBackoff(kind: Kind, delay: Delay | null | undefined, papieaDebug: boolean, retries: number = 0): Backoff {
```
Emulebest, line 77: This is too intrusive IMO, please take a look at my approach here https://github.com/nutanix/papiea/blob/e531a23927b9e66d4b7d93f05a824d9774332e57/papiea-engine/src/diff_resolver_main.ts#L49

Emulebest (changes requested):
I don't like passing debug variable to the business logic, please take a look at the example in the comments

Emulebest:
> > I don't think we should do this just in debug builds, I think it's quite relevant for release builds as well.
> 
> Yes, I am thinking about two things i.e.
> 
> 1. What would be an ideal value for this considering the latency and workflows for papiea, maybe @Emulebest can provide more inputs on this?
> 2. If (and how) we should allow the user to customize the value for the backoff exponent?

I'd rather let a user judge on the matters of the idea constant.

Anyhow, we could add it to one of our config variables like used here:
https://github.com/nutanix/papiea/blob/master/papiea-engine/papiea-config.yaml

https://github.com/nutanix/papiea/blob/master/papiea-engine/src/utils/arg_parser.ts

nitesh-idnani1:
> > > I don't think we should do this just in debug builds, I think it's quite relevant for release builds as well.
> > 
> > 
> > Yes, I am thinking about two things i.e.
> > 
> > 1. What would be an ideal value for this considering the latency and workflows for papiea, maybe @Emulebest can provide more inputs on this?
> > 2. If (and how) we should allow the user to customize the value for the backoff exponent?
> 
> I'd rather let a user judge on the matters of the idea constant.
> 
> Anyhow, we could add it to one of our config variables like it is used here:
> https://github.com/nutanix/papiea/blob/master/papiea-engine/papiea-config.yaml
> 
> https://github.com/nutanix/papiea/blob/master/papiea-engine/src/utils/arg_parser.ts

Added a config variable for the diff retry exponent value. Kindly check and review the changes. Thanks

papiea-engine/src/utils/utils.ts
```diff
@@ -125,8 +125,22 @@ export function deepMerge(target: any, ...sources: any[]): any {
     return deepMerge(target, ...sources);
 }
 
-export function calculateBackoff(retries: number, maximumBackoff: number, entropy: number) {
-    return Math.min(Math.pow(2, retries) + entropy, maximumBackoff)
+export function getCalculateBackoffFn(papieaDebug: boolean, retryExponent: number) {
+    let exponent: number
+    if (papieaDebug) {
+        console.log("Debug flag is set. Setting the diff retry exponent to 1.3")
```
Emulebest, line 131: Console.log is not a proper logging, please adjust to use the logger

papiea-engine/src/utils/utils.ts
```diff
@@ -125,8 +125,22 @@ export function deepMerge(target: any, ...sources: any[]): any {
...
+        exponent = 1.3
+    } else {
+        exponent = retryExponent
+    }
```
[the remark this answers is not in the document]
Emulebest: I second @joshua-berry-ntnx opinion, let's have it tunable. Moreover, this gets me to think whether we should maybe do the debug config instead of debug flags for such cases, @shlomi-nx @joshua-berry-ntnx what do you think?

Emulebest (changes requested):
One general caveat in the code. @shlomi-nx @joshua-berry-ntnx please take a look at the comment in the PR, any ideas?

shlomi-nx:
I too agree with hiding things behind `debug` flag is not good. But I think more generally, retry frequency should be related to the provider itself, not the engine. Some providers would need more rapid retries, while others not. 

We already have a per-handler retry mechanism, specified by the return value of the handler. Perhaps we could unblock ourselves there, while per-provider retry configuration is being worked on?

nitesh-idnani1:
Review and merge this PR, and I will start working on the discussed enhancement to add the diff retry in the kind registration. Thanks

shlomi-nx (changes requested):
I would still like to not use the general purpose `debug` flag. Lets replace it with `papiea_retry_exponent` to signal that this is a papiea-wide retry exponent. Later we could add provider/kind specific configuration as discussed.

nitesh-idnani1:
I'm ready with the new changes that we discussed i.e. allow user to specify kind-level diff retry exponent value. The changes are not big and compatible with this PR changes.
Let me know if I can add them here or create a separate PR for the new changes.

papiea-engine/src/utils/utils.ts
```diff
@@ -125,8 +126,15 @@ export function deepMerge(target: any, ...sources: any[]): any {
     return deepMerge(target, ...sources);
 }
 
-export function calculateBackoff(retries: number, maximumBackoff: number, entropy: number) {
-    return Math.min(Math.pow(2, retries) + entropy, maximumBackoff)
+export function getCalculateBackoffFn(retryExponent: number) {
+    return (retries?: number, maximumBackoff?: number, entropy?: number) => {
```
shlomi-nx, line 130: I like curry ;)

shlomi-nx (commented):
Looks good. Sure you can add the kind exponent configuration to this PR.

papiea-engine/src/utils/utils.ts
```diff
@@ -125,8 +126,19 @@ export function deepMerge(target: any, ...sources: any[]): any {
...
+        if (retries !== undefined && retries !== null &&
+            maximumBackoff !== undefined && maximumBackoff !== null &&
+            entropy !== undefined && entropy !== null) {
+            if (kind_retry_exponent !== null && kind_retry_exponent !== undefined) {
+                logger.info(`Retry exponent for kind ${ kind_name } is set, not using the config exponent value`)
+                retryExponent = kind_retry_exponent
+            }
+            return Math.min(Math.pow(retryExponent, retries) + entropy, maximumBackoff)
+        }
+        return 10
```
shlomi-nx, line 140: What is this 10 number?

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -170,7 +170,7 @@ export class ProviderSdk implements ProviderImpl {
         return this._server_manager.server;
     }
 
-    new_kind(entity_description: Data_Description): Kind_Builder {
+    new_kind(entity_description: Data_Description, diff_retry_exponent?: number): Kind_Builder {
```
shlomi-nx, line 173: Let's add this configuration as a function. 

```
const k = new kind(schema)
k.setExponentBackoff(1.2)
```

If we do want to keep it in the schema, we could add something like `x-papiea-backoff-exponent` tag that is allowed on a kind entity.

shlomi-nx (changes requested):
Looks good, but I rather not change the kind signature.

papiea-engine/src/utils/utils.ts
nitesh-idnani1: Added this for safety. If one of the param is null/undefined, set the backoff delay to 10 sec.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
nitesh-idnani1: Changing this to a function for now, we can change this later based on the requirements.

papiea-engine/src/utils/utils.ts
shlomi-nx: oh I see. Let's add a loud WARNING message there, to know that we are using unexpected defaults, and the reason we must have reached there.

shlomi-nx (approved):
Looks good to me

PR #615, opened 2021-02-04
Use kind structure in on_create if input schema is missing

### LINKED ISSUE #605: Enhancement: Custom constructors - if no schema is provided, use default spec schema
If the user wants to defines a custom constructor that needs to accept the kind's spec as input, he should be able to achieve that by simply omitting the schema from the on_create call.

nitesh-idnani1:
Closes #605 
- Changed input schema to optional parameter and use kind structure if it is undefined.
- Added test to verify the use of kind structure when input schema is undefined.

### DIFF
--- a/papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
+++ b/papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
@@ -604,11 +604,15 @@ export class Kind_Builder {
     }
 
     // Return type should always contain spec and status (metadata could be empty)
-    on_create(description: {input_schema: any, error_schemas?: ErrorSchemas}, handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<{spec: Spec, status: Status, metadata?: Partial<Metadata>}>): Kind_Builder {
+    on_create(description: {input_schema?: any, error_schemas?: ErrorSchemas}, handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<{spec: Spec, status: Status, metadata?: Partial<Metadata>}>): Kind_Builder {
         const name = `__${this.kind.name}_create`
         const loggerFactory = new LoggerFactory({logPath: name})
         const [logger, handle] = loggerFactory.createLogger()
         logger.info("You are registering on create handler. Note, this is a post create handler. The behaviour is due to change")
+        if (!description.input_schema) {
+            logger.info("Input schema for on_create is undefined, using the kind's entity structure")
+            description.input_schema = this.kind.kind_structure
+        }
         handle.cleanup()
         this.kind_procedure(name, description, handler)
         return this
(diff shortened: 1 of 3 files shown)


Emulebest (approved):

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -604,11 +604,15 @@ export class Kind_Builder {
...
 
     // Return type should always contain spec and status (metadata could be empty)
-    on_create(description: {input_schema: any, error_schemas?: ErrorSchemas}, handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<{spec: Spec, status: Status, metadata?: Partial<Metadata>}>): Kind_Builder {
+    on_create(description: {input_schema?: any, error_schemas?: ErrorSchemas}, handler: (ctx: ProceduralCtx_Interface, input: any) => Promise<{spec: Spec, status: Status, metadata?: Partial<Metadata>}>): Kind_Builder {
         const name = `__${this.kind.name}_create`
         const loggerFactory = new LoggerFactory({logPath: name})
         const [logger, handle] = loggerFactory.createLogger()
         logger.info("You are registering on create handler. Note, this is a post create handler. The behaviour is due to change")
+        if (!description.input_schema) {
+            logger.info("Input schema for on_create is undefined, using the kind's entity structure")
```
shlomi-nx, line 613: This error message does not say on which kind, provider or version. User will have no meaningful info.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -604,11 +604,15 @@ export class Kind_Builder {
...
+            description.input_schema = this.kind.kind_structure
```
shlomi-nx, line 614: Is this correct? doesn't `kind_structrure` contain status as well? This only needs to allow metadata and spec. Perhaps add a test that shows status is not added.

shlomi-nx (commented):
@Emulebest, @nitesh-idnani1, I know this is merged, but please see the comments.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
nitesh-idnani1: Yes, the kind structure for a kind is set to it's schema.
Reference: https://github.com/nutanix/papiea/blob/master/papiea-engine/__tests__/test_data/location_kind_test_data.yml

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
nitesh-idnani1: Will add the necessary information. Thanks for pointing this out

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
shlomi-nx: Of course it is set to it's schema, but it also contains status fields. Status fields must not be included in the constructor parameters by default. If the provider wants to expose anything other than spec+metadata, they must not use the default.

Please add a test that status fields are not allowed using the default constructor schema.

nitesh-idnani1: Yes, we already have a validator module to remove status fields in case of spec create/update.

shlomi-nx: Ok, that's good. Let's just add a test case that shows that status fields won't appear in constructors with default schema. Thanks!

PR #617, opened 2021-02-05
Added strict checking in validate status-only fields module in validator

### LINKED ISSUE #601: Validator is making assumptions on correctness that may not be true
Assumptions such as:
https://github.com/nutanix/papiea/blob/40ec741dd7a6890016696f3bba43189bf7caafae/papiea-engine/src/validator/index.ts#L240-L243

What if `field_schema["items"]` has no sub entry `type`? The user will get the following error:

```
papiea_1   | Default handle got error: Cannot read property 'includes' of undefined. 
```

We ~may want to~ should add some checks on these assumptions, and as a result spit out better and more precise error messages.

nitesh-idnani1:
Closes #601 
- Added value checks in validate_status_only_field function to avoid undefined value access in validator.
- Added test to verify the validator module.

### DIFF
--- a/papiea-engine/src/validator/index.ts
+++ b/papiea-engine/src/validator/index.ts
@@ -221,33 +221,37 @@ export class ValidatorImpl {
     }
 
     validate_status_only_field(schema: Data_Description) {
-        try{
+        try {
             for(let field in schema) {
                 const field_schema = schema[field]
-                if (field_schema["type"] === "object") {
-                    if (field_schema.hasOwnProperty("required") && field_schema.hasOwnProperty("properties")) {
-                        for (let req_field of field_schema["required"]) {
-                            if (field_schema["properties"][req_field].hasOwnProperty("x-papiea") && field_schema["properties"][req_field]["x-papiea"] === "status-only") {
-                                throw new ValidationError([{
-                                    name: "ValidationError",
-                                    message: `${req_field} of type 'status-only' is set to be required. Required fields cannot be 'status-only'`
-                                }])
+                if (field_schema.hasOwnProperty("type")) {
+                    if (field_schema["type"] === "object") {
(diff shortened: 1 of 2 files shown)


Emulebest (approved):

papiea-engine/src/validator/index.ts
```diff
@@ -221,33 +221,37 @@ export class ValidatorImpl {
...
-                                    message: `${req_field} of type 'status-only' is set to be required. Required fields cannot be 'status-only'`
-                                }])
+                if (field_schema.hasOwnProperty("type")) {
+                    if (field_schema["type"] === "object") {
+                        if (field_schema.hasOwnProperty("required") && field_schema.hasOwnProperty("properties")) {
+                            for (let req_field of field_schema["required"]) {
+                                if (field_schema["properties"][req_field].hasOwnProperty("x-papiea") && field_schema["properties"][req_field]["x-papiea"] === "status-only") {
+                                    throw new ValidationError([{
+                                        name: "ValidationError",
+                                        message: `${req_field} of type 'status-only' is set to be required. Required fields cannot be 'status-only'`
```
shlomi-nx, line 234: @Ntnx-Igor-Korotach, @nitesh-idnani1 , Message does not contain proper coordinates (provider, version, kind), and will be hard for users to identify,

shlomi-nx, line 234: @Ntnx-Igor-Korotach, @nitesh-idnani1 , Message does not contain proper coordinates (provider, version, kind), and will be hard for users to identify,

shlomi-nx (commented):
Please see below

papiea-engine/src/validator/index.ts
nitesh-idnani1: Will add the necessary information, thanks for pointing this out

PR #623, opened 2021-02-11
Improved logging information for papiea

### LINKED ISSUE #575: Add log tags and dynamic log levels

### LINKED ISSUE #577: Add log targets & querying
What we should also consider:

Log target options in config (in-memory, logstash, elasticsearch, cassandra among the popular)
Search utilities for log storages

### LINKED ISSUE #600: Errors often lack useful context
Some errors conveys no context, look at the following example:

```
papiea_1   | Error msg: Validation failed.. Details: ;
papiea_1   |   Cause: message - Error: hardware_id of type 'status-only' is set to be required. Required fields cannot be 'status-only'
```

When looking at the papiea log, how can we tell which provider, kind, version this error is related to? What if some kinds have more than a single `hardware_id` (perhaps show path from spec)? 

We should make sure we report useful context to the users, to help them identify where the problems are coming from.

This is just one example I found, but we should try and find more locations where we give errors without much context.

### LINKED ISSUE #619: Fixes & enhancements on the merged PRs
To do:

- [ ] https://github.com/nutanix/papiea/pull/617#pullrequestreview-585717095
- [ ] https://github.com/nutanix/papiea/pull/615#pullrequestreview-585712128
- [x] https://github.com/nutanix/papiea/pull/616#pullrequestreview-584433477

### LINKED ISSUE #628: Custom constructors: Validation errors are unclear [was: Mismatch between generated swagger and actual constructor arguments]
### System Details

- **Papiea Engine version**: 0.9.24
- **Papiea SDK version:** 0.9.24
- **Papiea Client version:** 0.9.24

### Describe the Bug

If I define a custom constructor with some arbitrary input parameters `{a, b}`, the generated swagger in the UI says I should be able to call the constructor with `a` and `b` parameters at the top level. Yet doing so leads to model validation errors.

It turns out that wrapping `a` and `b` inside a spec (like `{spec: {a: ..., b: ...}}` is required to call the constructor successfully, even though the swagger says this should not be required.

This extra `spec` field is also passed to the constructor function in the provider.

### Expected/Actual Behavior

I expect the swagger to agree with the actual behavior of the constructor—either everything should be wrapped in `spec`, or it shouldn't.

I'm not sure that wrapping everything in `spec` is desirable, since it's quite confusing to place parameters in `spec` which will eventually end up in an entity's `status` when the entity is actually created. Also, given that the constructor parameters might differ quite dramatically from the entity's actual `spec`, it seems misleading to present them as `spec` when they're really just arguments.

Either way, though, the swagger should be consistent with what Papiea actually accepts.

<hr>

nitesh-idnani1:
- Added request URL field in papiea error responses.
- Added relevant information in the error messages.
- Added helper module to transform validator library messages to relevant papiea messages.
- Added relevant information in the logger messages.
- Modified tests to verify the new logging changes.

Closes #575
Closes #577
Closes #600
Closes #619
Closes #628


papiea-client/src/entity_client.ts
```diff
@@ -193,7 +193,7 @@ async function wait_for_watcher_status(papiea_url: string, s2skey: string, watch
         const end_time: number = new Date().getTime()
         const time_elapsed = (end_time - start_time)/1000
         if (time_elapsed > timeout_secs) {
-            throw new Error("Timeout waiting for intent watcher status")
+            throw new Error("Timeout waiting for intent watcher status change")
```
shlomi-nx, line 196: Timeout on which entity? provider? kind? version? :)

papiea-engine/__tests__/entity_tests/extension.test.ts
```diff
@@ -123,7 +123,7 @@ describe("Entity API with metadata extension tests", () => {
             const res = err.response;
             expect(res.status).toEqual(400);
             expect(res.data.error.errors.length).toEqual(1);
-            expect(res.data.error.errors[0].message).toEqual('Metadata extension should be an object')
+            expect(res.data.error.errors[0].message).toContain('Metadata extension: \"\" should be an object for entity')
```
shlomi-nx, line 126: `for entity` conveys no useful information. Kind, provider, version do.

papiea-engine/__tests__/entity_tests/pagination.test.ts
```diff
@@ -155,7 +155,7 @@ describe("Pagination tests", () => {
                 }
             });
         } catch (e) {
-            expect(e.response.data.error.errors[0].message).toBe("Offset should not be less or equal to zero");
+            expect(e.response.data.error.errors[0].message).toBe("Offset should be greater than zero, received: 0");
```
shlomi-nx, line 158: I think not allowing offset of 0 is little too rigid. Offset of 0 is the same as no offset.

papiea-engine/__tests__/entity_tests/uuid_validation.test.ts
```diff
@@ -109,7 +109,7 @@ describe("Uuid validation tests", () => {
                 }
             })
         } catch (e) {
-            expect(e.response.data.error.message).toEqual("Uuid is not provided, but supposed to be since validation pattern is specified")
+            expect(e.response.data.error.message).toContain("Metadata uuid is undefined but validation pattern is specified for entity with kind")
```
shlomi-nx, line 112: Perhaps show the validation patten here. as well as kind, provider, version

papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
```diff
@@ -121,7 +121,7 @@ describe("Procedures tests", () => {
             const res = err.response;
             expect(res.status).toEqual(400);
             expect(res.data.error.errors.length).toEqual(1);
-            expect(res.data.error.errors[0].message).toBe("moveX with schema MoveInput was expecting non-empty input");
+            expect(res.data.error.errors[0].message).toBe("Procedure: moveX with schema: MoveInput was expecting non-empty input");
```
shlomi-nx, line 124: In which kind, provider, version?

papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
```diff
@@ -155,8 +155,8 @@ describe("Procedures tests", () => {
             expect(res.data.error.errors.length).toEqual(2);
             expect(res.data.error.message).toEqual("Procedure invocation failed.")
             expect(res.data.error.code).toEqual(500)
-            expect(res.data.error.errors[0].message).toEqual("x is a required field");
-            expect(res.data.error.errors[1].message).toEqual("y is a required field");
+            expect(res.data.error.errors[0].message).toContain("Input is missing required field: x");
```
shlomi-nx, line 158: For which procedure, on which kind, provider, version?

papiea-engine/src/auth/oauth2.ts
```diff
@@ -76,7 +76,7 @@ export function createOAuth2Router(logger: Logger, redirect_uri: string, provide
             await sessionKeyAPI.inactivateKey(sessionKey.key)
             await idpToken.revoke('access_token');
         } catch (e) {
-            logger.error('Logout error: ', e.message);
+            logger.error(`Failed to logout user with token: ${token} due to error: ${e.message}`);
```
shlomi-nx, line 79: On which provider? version?

papiea-engine/src/auth/authz.ts
```diff
@@ -159,32 +159,32 @@ export class AdminAuthorizer extends Authorizer {
...
-                throw new PermissionDeniedError();
+                throw new PermissionDeniedError(`User has wrong owner for the object's user info with owner: ${object.owner} in admin permission check to create s2skey`);
             }
             return;
         }
         if (action === Action.ReadS2SKey || action === Action.InactivateS2SKey) {
             if (object.owner !== user.owner
                 || (user.provider_prefix !== undefined && object.provider_prefix !== user.provider_prefix)) {
-                throw new PermissionDeniedError();
+                throw new PermissionDeniedError(`User has wrong provider prefix for the object with prefix: ${object.provider_prefix} in admin permission check to read/inactivate s2skey`);
```
shlomi-nx, line 180: Don't have to mention `prefix` in the message. Provider prefix is just the provider name..

papiea-engine/src/auth/session_key.ts
```diff
@@ -64,7 +64,7 @@ export class SessionKeyAPI {
             sessionKey.idpToken = accessToken
             return sessionKey
         } catch (e) {
-            throw new Error(`Couldn't refresh the token: ${e.message}`)
+            throw new Error(`Couldn't refresh the session token: ${e.message}`)
```
shlomi-nx, line 67: Session for which provider? which idp url? maybe which user?

papiea-engine/src/databases/graveyard_db_mongo.ts
```diff
@@ -46,10 +46,10 @@ export class Graveyard_DB_Mongo implements Graveyard_DB {
                 "metadata.provider_version": entity_ref.provider_version
             })
         if (result.result.n === undefined || result.result.ok !== 1) {
-            throw new Error("Failed to remove entity");
+            throw new Error(`MongoDBError: Failed to remove entity with uuid: ${entity_ref.uuid}, kind: ${entity_ref.kind}, provider_prefix: ${entity_ref.provider_prefix}, provider_version: ${entity_ref.provider_version}`);
```
shlomi-nx, line 49: You could simply write `... provider: ${entity_ref.provider_prefix}/${entity_ref.provider_version} ..` when you want to log the provider information. This way it will look like `blah blah on provider shlomi/1.0.1`. 

This holds for all the logs, makes it easier to read

papiea-engine/src/databases/intent_watcher_db_mongo.ts
```diff
@@ -34,7 +35,7 @@ export class IntentWatcher_DB_Mongo implements IntentWatcher_DB {
             "uuid": uuid,
         });
         if (result === null) {
-            throw new Error("key not found");
+            throw new EntityNotFoundError('IntentWatcher', uuid);
```
shlomi-nx, line 38: IntentWatcher uuids are nice, but I would like an explicit print of the provider/version/kind and even entity uuid in  logs that have intentwatcher in them.

papiea-engine/src/intentful_core/intentful_context.ts
```diff
@@ -56,7 +56,7 @@ export class IntentfulContext {
     getIntentfulStrategy(kind: Kind, user: UserAuthInfo): IntentfulStrategy {
         const strategy = this.behaviourStrategyMap.get(kind.intentful_behaviour)
         if (strategy === undefined) {
-            throw new Error(`Strategy associated with behaviour: ${kind.intentful_behaviour} not found`)
+            throw new Error(`Strategy associated with behaviour: ${kind.intentful_behaviour} not found for kind: ${kind.name}`)
```
shlomi-nx, line 59: on `provider/version`

papiea-engine/src/intentful_engine/diff_resolver.ts
```diff
@@ -114,7 +114,7 @@ export class DiffResolver {
                 metadata, provider, kind, spec, status,
             };
         } catch (e) {
-            this.logger.debug(`Couldn't rediff entity with uuid ${entry_reference.entity_reference.uuid}: ${e}. Removing from watchlist`)
+            this.logger.debug(`Couldn't generate diff for entity with uuid: ${entry_reference.entity_reference.uuid} and kind: ${entry_reference.entity_reference.kind} due to error: ${e}. Removing from watchlist`)
```
shlomi-nx, line 117: `Couldn't generate diff for entity {uuid} on {provider}/{version}/{kind} due to ${e} ...` 

Also, note that `${e}` does not print the stack trace, so we are swallowing the stack trace here, which is very unfortunate.

shlomi-nx (changes requested):
Reviewed most files but not all changes.

Please see my comments below, I still see some context missing, and added suggestions for improved log format. Also please be sure we do not swallow stack traces.

Emulebest (changes requested):
I second Shlomi's review. I'd suggest extending or changing PapieaError's constructor for it to optionally include all the provider and entity info to extract

shlomi-nx:
+1 on @Ntnx-Igor-Korotach's idea, if we formalize the entity information for exceptions and logs we could achieve two things:
1. Uniformity - all logs will have the same info without the need to remember manually to write them
2. Searchability - If these would get logged as json entries, we could later filter logs of particular entities/providers etc.

papiea-backend-utils/src/index.ts
```diff
@@ -17,3 +18,71 @@ export {
     getTracer, getTracingMiddleware, spanOperation, spanEntityOperation,
     RequestContext, getTraceHeaders
 };
+
+export class EntityLoggingInfo {
```
Emulebest, line 22: This class doesn't solve the problem of uniformity @shlomi-nx and I raised since you essentially need to remember to pass it as a second parameter. This is good as an aggregator class but what you actually need is a strict error interface that requires something like this in the constructor (possibly using this class). 

Also `setAdditionalField` is a good fit for Java but there is no need to pollute the environment with these calls if you can pass an `any` type parameter in the constructor of factory function.

`toString` should be called internally from the single entry point (for reference you can use https://github.com/nutanix/papiea/blob/78435c935e6b5f5d021ac46ec0193d8611ba5703/papiea-engine/src/main.ts#L95)

Emulebest (changes requested):

papiea-backend-utils/src/index.ts
nitesh-idnani1: Having it in the constructor would still require us to initialize it every time we throw an error right?

The idea behind `setAdditionalField` is to simplify the logging interface (similar to the sdk interface).

I'll work on the toString part and add in the papiea error implementation. Thanks for pointing that out.

papiea-engine/src/databases/utils/errors.ts
```diff
@@ -1,38 +1,48 @@
...
 
-    constructor(kind: string, uuid: uuid4) {
+    constructor(kind: string, uuid: uuid4, provider_prefix: string, provider_version: string) {
         super();
         this.kind = kind;
         this.uuid = uuid;
+        const additional_info = { "entity_uuid": uuid }
+        this.entity_info = new EntityLoggingInfo(provider_prefix, provider_version, kind, additional_info)
     }
 }
```
Emulebest, line 48: This is a good place to highlight what should be changed.

In the comments above I spoke about `PapieaError` being implemented. These are my thought on how that should be done:

1) Make `PapieaError` (or better `PapieaException` to not collide with the existing naming) inherit `Error`
2) `PapieaException` should have a constructor that has an `exception_context` essential argument that features all the possible options (`prefix`, `version`, `kind`) as optional properties (this should be typed out as `PapieaExceptionContext`)
3) Use `EntityLoggingInfo` internally to aggregate those as good error strings
4) Override the `toString` function for generic view of error context

Then working with classes like `EntityNotFoundError`:

1) Inherit from `PapieaException`
2) Have signature feature essential param needed in the super class constructor
e.g.
```
constructor(kind: string, uuid: uuid4, exception_context: PapieaExceptionContext)
```
3) Override `toString` to add up messages on top of `PapieaException`'s `toString`
4) Change https://github.com/nutanix/papiea/blob/78435c935e6b5f5d021ac46ec0193d8611ba5703/papiea-engine/src/errors/papiea_error_impl.ts#L63 to use errors' `toString` instead of `err.message`

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -310,15 +310,17 @@ export class Entity_API_Impl implements Entity_API {
...
             }
         } else {
             const [metadata, _] = await this.spec_db.get_spec(entityRef);
             const has_perm = await this.has_permission(user, provider, metadata, action)
             if (has_perm) {
                 return {"success": "Ok"}
             } else {
-                throw new PermissionDeniedError(`User: ${JSON.stringify(user)} does not have ${action} permission on entity with uuid: ${entityRef.uuid} and kind: ${entityRef.kind} in provider with prefix: ${provider.prefix} and version: ${provider.version}`)
+                const additional_info = { "entity_uuid": entityRef.uuid, "user": JSON.stringify(user), "action": action }
+                throw new PermissionDeniedError(`User does not have permission on entity`, entityRef.provider_prefix, entityRef.provider_version, entityRef.kind, additional_info)
```
Emulebest, line 323: In place additional info when properly formatted is better for readability

Emulebest (changes requested):
Looks better but please see the comments :)

papiea-engine/src/errors/validation_error.ts
```diff
@@ -1,14 +1,19 @@
...
+        super(JSON.stringify(messages), context);
         Object.setPrototypeOf(this, ValidationError.prototype);
         this.errors = messages;
-        this.entity_info = new EntityLoggingInfo(provider_prefix, provider_version, kind_name, additional_info);
+    }
+
+    toErrors(): { [key: string]: any }[] {
+        return this.errors.map(description => {
+            return { message: description }
+        });
```
Emulebest, line 17: Does this include formatted messages from `PapieaException` base class?

Emulebest (commented):
Looks good! @nitesh-idnani1 is this completely ready or you are still working on it?

papiea-engine/src/errors/validation_error.ts
nitesh-idnani1: No, since the papiea base class contains only the exception context I've separated its usage in papiea error impl like this
https://github.com/nutanix/papiea/blob/934c69fcb68d67165228b4b2aae8c3bce3b8d255/papiea-engine/src/errors/papiea_error_impl.ts#L33

papiea-client/src/entity_client.ts
```diff
@@ -193,7 +196,7 @@ async function wait_for_watcher_status(papiea_url: string, s2skey: string, watch
         const end_time: number = new Date().getTime()
         const time_elapsed = (end_time - start_time)/1000
         if (time_elapsed > timeout_secs) {
-            throw new Error("Timeout waiting for intent watcher status")
+            throw new Error(`Timeout waiting for intent watcher status change with uuid: ${watcher.uuid} and entity uuid: ${watcher.entity_ref.uuid}`)
```
shlomi-nx, line 199: provider/version/kind

papiea-engine/__tests__/entity_tests/uuid_validation.test.ts
```diff
@@ -109,7 +109,7 @@ describe("Uuid validation tests", () => {
                 }
             })
         } catch (e) {
-            expect(e.response.data.error.message).toEqual("Uuid is not provided, but supposed to be since validation pattern is specified")
+            expect(e.response.data.error.errors[0].message).toBe(`Metadata uuid is undefined but kind has validation pattern set`)
```
shlomi-nx, line 112: provider/version/kind. Let's also show the pattern. 
Perhaps its useful to also add the context of the call, something like 
```
Creating a new entity of {provider/version/kind} failed. Metadata uuid received is undefined but the kind has the following unsatisfied validation pattern {pattern}
```

papiea-engine/src/intentful_core/sfs_compiler.ts
```diff
@@ -16,7 +17,7 @@ export class SFSCompiler {
             sfs_parser(signature)
         } catch(e) {
             throw new ValidationError([
-                new Error(`SFS: '${signature}' parsing on kind: ${kind_name} failed with error: ${e.message}`)
+                new PapieaException(`SFS parsing on kind failed with error: ${e.message}`, { kind_name: kind_name, additional_info: { "signature": signature }})
```
shlomi-nx, line 20: Please add kind coordinates {provider/version/kind}. Perhaps rename `signature` to SFS, to be consistent with the message. 

Also notice that the stacktrace is swallowed, lets find a way to always aggregate stacktraces, never drop them.

papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
```diff
@@ -121,7 +121,7 @@ describe("Procedures tests", () => {
             const res = err.response;
             expect(res.status).toEqual(400);
             expect(res.data.error.errors.length).toEqual(1);
-            expect(res.data.error.errors[0].message).toBe("moveX with schema MoveInput was expecting non-empty input");
+            expect(res.data.error.errors[0].message).toBe(`Procedure was expecting non-empty object`);
```
shlomi-nx, line 124: Kind coordinates are missing from the error message. Also some context:
```
Could not invoke procedure {provider/version/kind/procedure}. Procedure definition expects a non-empty object, but an empty object was provided.
```

papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
```diff
@@ -155,8 +155,8 @@ describe("Procedures tests", () => {
             expect(res.data.error.errors.length).toEqual(2);
             expect(res.data.error.message).toEqual("Procedure invocation failed.")
             expect(res.data.error.code).toEqual(500)
-            expect(res.data.error.errors[0].message).toEqual("x is a required field");
-            expect(res.data.error.errors[1].message).toEqual("y is a required field");
+            expect(res.data.error.errors[0].message).toContain(`Input is missing required field: x`);
```
shlomi-nx, line 158: I may be missing the full message, but I think it should be something like:

```
Could not invoke procedure {provider/version/kind/procedure}. Received procedure arguments is missing a required field {x}.
```

papiea-engine/__tests__/provider_tests/provider_api.test.ts
```diff
@@ -82,7 +82,7 @@ describe("Provider API tests", () => {
         try {
             await providerApi.post('/', provider);
         } catch (e) {
-            expect(e.response.data.error.errors[0].message).toBe("x-papiea has wrong value: spec-only, possible values are: status-only");
+            expect(e.response.data.error.errors[0].message).toBe("x-papiea has wrong value: spec-only, correct values include: status-only");
```
shlomi-nx, line 85:
```
Cannot register {provider}/{version}. Kind {kind} has invalid definition: x-papiea has wrong value: spec-only, correct values are: status-only
```

papiea-engine/__tests__/provider_tests/provider_api.test.ts
```diff
@@ -252,7 +252,7 @@ describe("Provider API tests", () => {
         } catch (e) {
             expect(e).toBeDefined()
             expect(e.response.data.error.message).toEqual("Validation failed.")
-            expect(e.response.data.error.errors[0].message).toEqual("Status body is undefined, please use null fields instead")
+            expect(e.response.data.error.errors[0].message).toEqual(`Status body is undefined for entity, please use null fields instead`)
```
shlomi-nx, line 255:
```
Cannot update status for {provider/version/kind}. Status body contains undefined fields which Papiea does not support. If you are trying to remove a field from status it should be set to null value instead.
```

papiea-engine/src/auth/casbin.ts
```diff
@@ -31,11 +33,11 @@ export class CasbinAuthorizer extends Authorizer {
     async checkPermission(user: UserAuthInfo, object: any, action: Action): Promise<void> {
         try {
             if (!this.enforcer.enforce(user, object, action)) {
-                throw new PermissionDeniedError();
+                throw new PermissionDeniedError(`User does not have permission for the entity`, { additional_info: { "user": JSON.stringify(user), "action": action, "entity": JSON.stringify(object) }});
```
shlomi-nx, line 36: Missing provider/version

papiea-engine/src/auth/oauth2.ts
```diff
@@ -70,13 +70,13 @@ export function createOAuth2Router(logger: Logger, redirect_uri: string, provide
...
         const token = req.user.authorization.split(' ')[1]
-        const sessionKey = await sessionKeyAPI.getKey(token, oauth2)
+        const sessionKey = await sessionKeyAPI.getKey(token, oauth2, provider.prefix, provider.version)
         const idpToken = oauth2.accessToken.create({ "access_token": sessionKey.idpToken.token.access_token });
         try {
             await sessionKeyAPI.inactivateKey(sessionKey.key)
             await idpToken.revoke('access_token');
         } catch (e) {
-            logger.error('Logout error: ', e.message);
+            logger.error(`Failed to logout user with token: ${token} in provider with prefix: ${provider.prefix} and version: ${provider.version} due to error: ${e.message}`);
```
shlomi-nx, line 79: Can we condense the provider/prefix to simply that? 

```
Failed to logout user with token: ${token} for ${provider.prefix}/${provider.version} due to error: ${e.message}
```

papiea-engine/src/databases/intent_watcher_db_mongo.ts
shlomi-nx: This is resolved, but I am not seeing the comment addressed here.. Am I missing something?

papiea-engine/src/entity/entity_api_impl.ts
```diff
@@ -185,16 +186,17 @@ export class Entity_API_Impl implements Entity_API {
             user, prefix, version, kind_name, entity_uuid, ctx);
         const procedure: Procedural_Signature | undefined = kind.entity_procedures[procedure_name];
         if (procedure === undefined) {
-            throw new Error(`Procedure ${procedure_name} not found for kind ${kind.name}`);
+            throw new PapieaException(`Entity procedure not found`, { provider_prefix: prefix, provider_version: version, kind_name: kind.name, additional_info: { "procedure_name": procedure_name }});
```
shlomi-nx, line 189: I see now that perhaps some of the comments I added above about missing information actually have the info supplied in the structured fields and additional_info. However, maybe it's worthwhile to still keep a nice, readable string along with the structure. Here for example I would write:

```
Entity procedure {procedure_name} not found on {provider/version/kind}.
```

shlomi-nx (commented):
Reviewed partially, that's a loooong PR! :)

I added a bunch of improved error texts. I later saw that errors contain structured fields, but I thought that having a meaningful string can be helpful when trying to read the log. 

Let me know what you think about my proposals and lets try to get this merged asap. This looks very good, and will be a wonderful improvement for our users!

papiea-engine/__tests__/entity_tests/uuid_validation.test.ts
nitesh-idnani1: Yes, that info is added in the error context when we throw this error on the engine side

papiea-engine/src/intentful_core/sfs_compiler.ts
nitesh-idnani1: @Emulebest 
We should discuss how to retain the stack trace for the errors. I think it'll involve changes in our error implementations.

papiea-engine/src/databases/intent_watcher_db_mongo.ts
nitesh-idnani1: Yes, I believe intent watchers do not have the provider related information in their context since they are not associated with one. The only way to add provider information is to fetch intent watcher and use its entity referenced provider info.

Here, since we're not able to fetch the watcher itself there's no way to associate the entity error with a particular provider. However, I've added provider context at other places where we have a valid intent watcher.

papiea-engine/src/entity/entity_api_impl.ts
nitesh-idnani1: I tried that earlier but I think it's better for readability and debugging purposes to have a clear concise error message separated from the context (entity-related) information.
It can help the end user to get a clear understanding of what went wrong and with which entity did it happen, rather than having all the information cluttered together (quite a few of the papiea messages are already long enough for that).

Emulebest: Agree with Nitesh, moreover it is important to get the context of the message to the user. Good looking error string becomes a burden and code mess, not even touching the point of needing for some reason to change the format of `on {provider/version/kind}`. Additionally, some places don't feature one of the components (no kind for example) which makes it a bit hard to keep consistent string

papiea-engine/src/intentful_core/sfs_compiler.ts
Emulebest: The error stacktraces are stored in `error.stack`. One possible solution is keeping a tree-like structure of error references so that in the `PapieaErrorResponseImpl` we can traverse, aggregate and log the stacktraces. This could be achieve by providing optional 3rd or 2nd argument where error nesting is due or by traversing the array of nested errors in case of `ValidationError`

papiea-engine/src/errors/validation_error.ts
Emulebest: Sure, well enough

papiea-engine/src/intentful_core/sfs_compiler.ts
nitesh-idnani1: I don't see the stack trace getting swallowed in cases of nested errors. I tried to print the stack trace for the above statement and got the following in the logs:
```
papiea-engine    | Error: ["SFS parsing on kind failed with error: Parse error at line 1, column 6:\nwrong, wrong2\n     ^\nExpected:\n\".\"\n\n"]
papiea-engine    |     at Function.try_parse_sfs (/code/papiea-engine/src/intentful_core/sfs_compiler.ts:21:19)
papiea-engine    |     at /code/papiea-engine/src/validator/index.ts:194:70
papiea-engine    |     at Array.forEach (<anonymous>)
papiea-engine    |     at ValidatorImpl.validate_sfs (/code/papiea-engine/src/validator/index.ts:194:43)
papiea-engine    |     at Provider_API_Impl.register_provider (/code/papiea-engine/src/provider/provider_api_impl.ts:48:24)
papiea-engine    |     at processTicksAndRejections (internal/process/task_queues.js:97:5)
papiea-engine    |     at /code/papiea-engine/src/provider/provider_routes.ts:11:24
```
Maybe we can add some field in the `PapieaErrorResponseImpl` to ensure we get a valid stack trace for all the error types

papiea-engine/src/entity/entity_api_impl.ts
shlomi-nx: I meant to have both an easily readable string where the information is available, and the structured error information.

PR #624, opened 2021-02-11
Fix client tracing issues

### LINKED ISSUE #620: Remove jaeger logs and fix closing
Jaeger is spitting too many logs

Emulebest:
resolves #620


shlomi-nx (approved):
Looks good. So we still have a close method, but we don't have to call it on every client instance, right?

Emulebest:
> Looks good. So we still have a close method, but we don't have to call it on every client instance, right?

Correct

PR #630, opened 2021-02-12
Add argument validation

### LINKED ISSUE #627: logging_verbosity and tracing_config config entities cannot be set thru env variables and have inadequate validation
### System Details

- **Papiea Engine version**: master e70375cf (0.9.24)
- **Papiea SDK version:** 0.9.24
- **Papiea Client version:** 0.9.24

### Describe the Bug

There are two very closely related problems:
- It is impossible to set the `logging_verbosity` or `tracing_config` parameters through the environment
- Neither of these options appear to have adequate validation, so setting either of them to an invalid value causes all kinds of weird behavior

### Bug information

From looking at `papiea-engine/utils/arg_parser.ts`, I see that:

```ts
const id: (val: any) => any = (val: any) => JSON.parse(JSON.stringify(val))

const TRANSFORM_FN_MAP: ... = {
...
    logging_verbosity: id,
    tracing_config: id
}
```

But this is clearly wrong as a matter of type safety—these are typed/complex JS values, not strings, yet `JSON.parse` returns a string. I don't think `id` is needed and they should actually just be `JSON.parse`d.

However, this also raises the matter of validation, as setting either of these variables via the environment results in them being set to strings. I tried hacking in JSON.parse in place of `id` as described above, but accidentally set the wrong value for logging_verbosity, and then instead of actual audit logs I got a lot of:

```
error    2021-02-12T01:11:12.638Z TypeError: Cannot read property 'includes' of undefined
    at AuditLogger.log (/code/papiea-engine/build/audit_logging.js:42:31)
        -> /code/papiea-engine/src/audit_logging.ts:64:31
    at ServerResponse.<anonymous> (/code/papiea-engine/build/audit_logging.js:24:22)
        -> /code/papiea-engine/src/audit_logging.ts:45:22
    at ServerResponse.emit (events.js:326:22)
    at onFinish (_http_outgoing.js:735:10)
    at onCorkedFinish (_stream_writable.js:673:5)
    at afterWrite (_stream_writable.js:490:5)
    at afterWriteTick (_stream_writable.js:477:10)
    at processTicksAndRejections (internal/process/task_queues.js:83:21)
```

### Expected/Actual Behavior

- Config inputs should be validated, and if the config is invalid, papiea-engine should refuse to start with an appropriate error
- Parse complex config values that come in through the environment properly

### Impact
Unable to change any Papiea logging settings, since the environment variables cannot be set correctly. (We cannot use the config file because Papiea is currently being configured thru a docker-compose file.)

<hr>

Emulebest:
resolves #627


joshua-berry-ntnx (commented):
I'll leave it to @shlomi-nx or @nitesh-idnani1 to give approval, but from looking over it quickly, it does seem like it addresses the issue. Thanks @Emulebest!

papiea-engine/src/utils/arg_parser.ts
```diff
@@ -123,6 +165,9 @@ export function getConfig(): PapieaConfig {
     for (let key in PAPIEA_DEFAULT_CFG) {
         if (PAPIEA_DEFAULT_CFG.hasOwnProperty(key) && !config.hasOwnProperty(key)) {
```
nitesh-idnani1, line 166: We don't need the first condition in the 'if' check. Since we're iterating through the default config, `hasOwnProperty` will always return true.

nitesh-idnani1:
Also, I notice this test is failing since the expect limit is lower than the time diff resolver gets invoked (not related to your changes)
https://github.com/nutanix/papiea/blob/2e52e8ecf04447fa498b76de3ff2b65f8dc98614/papiea-sdk/typescript/__tests__/intentful_workflow_tests/intentful_workflow.test.ts#L323

Maybe, you can change this to 7 or 8 to be sure or change the expect condition

papiea-engine/src/utils/arg_parser.ts
Emulebest: When iterating with `let key in ...`, if the object has a non basic prototype then it will inherit all its members (https://stackoverflow.com/questions/32422867/when-do-i-need-to-use-hasownproperty/32422910). Thus, this is a defensive mean to avoid theoretical security vulnerabilities.

PR #635, opened 2021-02-23
Background task implementation

### LINKED ISSUE #582: Background/periodic task scheduling for providers
### Problem Description

There are some (perhaps many) situations where providers will need to periodically and/or proactively poll external resources to understand the current state of the world. For example, a provider which manages IoT devices may wish to poll each device to determine its health status, and then expose this information through the corresponding entity's `status` fields. Or the provider may need to periodically poll an external list of devices to determine whether any have been added or removed (such that corresponding entities can be created or deleted).

### Expected/Actual Behavior
For now I think it would be sufficient for a provider to be able to register some type of periodic task with the engine. Any such task could be run by the engine using a mechanism similar to what is done with intent handlers—that is, a single instance of a provider is chosen to run the task, and the engine monitors the task and restarts it if the instance running the task has failed. (And of course the task itself can fail, but I don't have a strong opinion on what happens then. I think it's OK not to restart it—it's already periodic anyway—or perhaps have a configurable policy for what to do in that case.)

Later, some form of work partitioning might be nice, e.g. "for all entities of kind X, periodically run task Y", and have the work monitored and appropriately divided between all instances of a provider. But I don't see an immediate need for this (and I don't think we need to go full MapReduce just yet ;) ). Just mentioning it here as something to keep in the back of our minds.

### Impact
In a non-distributed world, this could be handled by the providers themselves—e.g. through the use of interval timers, cron jobs, etc. However, providers are supposed to be stateless and horizontally scalable, and any such background task necessarily involves maintaining some state if it is long-running.

Without this feature, instances of a provider would have to coordinate amongst themselves to ensure work was not being duplicated and sub-tasks were being partitioned appropriately. One workaround might be to abuse an intent handler—for example, create a special entity with a spec/status `token` field. Have each instance try to periodically change the token's spec such that there is a diff whenever the task should be run. Then the token's intent handler would be run in a single instance chosen by the engine. The intent handler could run the task and when the task is complete, set the `status.token = spec.token`. (Of course, this assumes it's possible for a provider to proactively make calls back to the engine, which I think is possible but haven't looked into.)

<hr>

Emulebest:
resolves #582

### DIFF
--- a/papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
+++ b/papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
@@ -36,7 +36,6 @@ import {getTracer, LoggerFactory} from "papiea-backend-utils"
 import {InvocationError, SecurityApiError} from "./typescript_sdk_exceptions"
 import {get_papiea_version, spanSdkOperation, validate_error_codes} from "./typescript_sdk_utils"
 import {Tracer} from "opentracing"
-import uuid = require("uuid")
 
 class SecurityApiImpl implements SecurityApi {
     readonly provider: ProviderSdk;
@@ -276,8 +275,8 @@ export class ProviderSdk implements ProviderImpl {
         return this
     }
 
-    async background_task(callback: () => Promise<any>, timeout_sec: number, name: string, schema?: any): Promise<BackgroundTaskBuilder> {
-        return await BackgroundTaskBuilder.create_task(this, callback, timeout_sec, this._tracer, name, schema)
+    background_task(name: string, delay_sec: number, callback: () => Promise<any>, schema?: any): BackgroundTaskBuilder {
+        return BackgroundTaskBuilder.create_task(this, name, delay_sec, callback, this._tracer, schema)
     }
 
     async register(): Promise<void> {
@@ -447,10 +446,10 @@ export class BackgroundTaskBuilder {
     private kind_client: EntityCRUD
     static BackgroundTaskState = class {
         static RunningSpecState() {
-            return "Running"
+            return "Should Run"
         }
         static RunningStatusState() {
-            return "To Be Run"
+            return "Running"
         }
         static IdleSpecState() {
             return "Idle"
@@ -469,10 +468,11 @@ export class BackgroundTaskBuilder {
         this.kind_client = kind_client(provider.papiea_url, provider.get_prefix(), name, provider.get_version(), provider.s2s_key)
     }
 
-    static async create_task(provider: ProviderSdk, callback: (entity?: Entity) => Promise<any>, timeout_sec: number, tracer: Tracer, name: string, schema?: any) {
-        const kind = provider.new_kind(schema ? this.modify_task_schema(schema) : {
+    static create_task(provider: ProviderSdk, name: string, delay_sec: number, callback: (entity?: Entity) => Promise<any>, tracer: Tracer, schema?: any): BackgroundTaskBuilder {
+        const kind = provider.new_kind(schema ? this.modify_task_schema({[name]: schema}) : {
             [name]: {
                 type: "object",
+                "x-papiea-entity": "differ",
                 properties: {
                     state: {
                         type: "string"
@@ -483,10 +483,16 @@ export class BackgroundTaskBuilder {
         kind.on("state", async (ctx, entity, input) => {
             await callback(entity)
             return {
-                delay_secs: timeout_sec
+                delay_secs: delay_sec
             }
         })
         return new BackgroundTaskBuilder(provider, tracer, name, kind)
+    }
+
+    private async update_task_entity() {
+        if (this.task_entity) {
+            this.task_entity = await this.kind_client.get(this.task_entity.metadata)
+        }
     }
 
     async start_task() {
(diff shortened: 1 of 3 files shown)


papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -271,6 +276,10 @@ export class ProviderSdk implements ProviderImpl {
         return this
     }
 
+    async background_task(callback: () => Promise<any>, timeout_sec: number, name: string, schema?: any): Promise<BackgroundTaskBuilder> {
```
shlomi-nx, line 279: I'd change the order of the parameters: name, delay_sec (can this be fractional seconds?), callback, schema?)

I only put schema last because it is not mandatory. Perhaps we could remove schema from this function and add another function such as `background_task_with_schema` that has schema before the callback.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -428,6 +437,129 @@ class Provider_Server_Manager {
...
+    private name: string
+    private kind: Kind_Builder
+    private task_entity: Entity | null = null
+    private kind_client: EntityCRUD
+    static BackgroundTaskState = class {
+        static RunningSpecState() {
+            return "Running"
+        }
+        static RunningStatusState() {
+            return "To Be Run"
```
shlomi-nx, line 453: Minor: the strings looks reversed. Spec, which is the desire should be "Should Run" and status, which is current state, should be "Running"

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -428,6 +437,129 @@ class Provider_Server_Manager {
...
+                }
+            });
+        }
+    }
+
+    async stop_task() {
+        const states = BackgroundTaskBuilder.BackgroundTaskState
+        if (this.task_entity === null) {
+            throw new Error(`Attempting to stop missing background task on provider: 
+                            ${this.provider.get_prefix()}, ${this.provider.get_version()}`)
```
shlomi-nx, line 517: We should also print the task name

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -428,6 +437,129 @@ class Provider_Server_Manager {
...
+        if (this.task_entity === null) {
+            this.task_entity = await this.kind_client.create({spec: {state: states.RunningSpecState()}})
+            await this.provider.provider_api_axios.patch(`${this.provider.provider_url}/${this.provider.get_prefix()}/${this.provider.get_version()}/update_status`,{
+                entity_ref: this.task_entity,
+                status: {
+                    state: states.RunningStatusState()
+                }
+            });
+        } else {
+            await this.kind_client.update(this.task_entity.metadata, {state: states.RunningSpecState()})
```
shlomi-nx, line 503: This seems like it may be a bug. We only capture `this.task_entity` once, which means it's `metadata.spec_version` will not increase as we stop and start the task. We should probably first `get` the entity to be on the correct spec_version.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -428,6 +437,129 @@ class Provider_Server_Manager {
...
+        const kind = provider.new_kind(schema ? this.modify_task_schema(schema) : {
+            [name]: {
+                type: "object",
+                properties: {
+                    state: {
+                        type: "string"
+                    }
+                }
+            }
+        })
```
shlomi-nx, line 482: This seems wrong, it either takes the custom schema, or the task's schema. This should be:
```typescript
const kind = provider.new_kind({
          name: {
                type: "object",
                properties: {
                    state: {  type: "string"},
                 
                    // All here should be status-only
                    // More generic task stuff could be here (TBD), e.g.: 
                    times_called: { type: number},
                    last_error: {string},
                    last_execution_time_ms: {number},
                    average_execution_time_ms: {number},
                    etc..

                    // And for custom schema:
                    custom: schema ? this.modify_task_schema(schema) : {}  // Or simply remove custom 
                }
            }
})
```

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -428,6 +437,129 @@ class Provider_Server_Manager {
...
+            throw new Error(`Attempting to update missing background task on provider: 
+                            ${this.provider.get_prefix()}, ${this.provider.get_version()}`)
+        } else {
+            await this.provider.provider_api_axios.patch(`${this.provider.provider_url}/${this.provider.get_prefix()}/${this.provider.get_version()}/update_status`,{
+                entity_ref: this.task_entity,
+                status: {state: this.task_entity.status.state, ...status}
+            });
+        }
+    }
+}
```
shlomi-nx, line 561: Who can use this function? If it's an API for the provider writer, this should only allow changing the `status.custom` field based on the custom schema provided.

shlomi-nx (commented):
Looks good, see my comments below. I'd like to also get @joshua-berry-ntnx's opinion on this.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
shlomi-nx: We can even split the background task type into two types, one for standard task without a custom schema, which would not even expose `update_status` function, and one with a custom schema that does.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
Emulebest: Let's talk on that, I am not sure this is a good idea

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
Emulebest: Once again, let's discuss this, I am not sure that having complex interfaces for this is a very good idea

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
Emulebest: Yep, thought about that, will adjust

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
Emulebest: Sure

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
Emulebest: Sounds good!

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
shlomi-nx: Sure, we'll talk about this on our meeting today.

PR #636, opened 2021-02-23
Fix error details swallowing in papiea

### LINKED ISSUE #633: Propagate errors with complete information
### System Details

- **Papiea version**: 0.9.27 [all components]

### Describe the Bug

There are a few situations where relevant error information is swallowed because exceptions are being re-thrown or are not being translated correctly across API boundaries. Please take a systematic look at all the places where we are re-wrapping exceptions and either stop doing this entirely or wrap them with the full context/info from the original error.

### Bug information
There are a couple places where we need to take special care:

**Custom Constructors:** Errors are swallowed/wrapped with some error that looks like "On Create couldn't be called; ..." and the original error details (except for the message) are lost. This "On Create couldn't be called" message is redundant; presumably the caller knows they are calling a constructor.

**Procedure Invocations:** Here again we have some wrapping which adds a redundant "Procedure invocation failed" message and again details are lost in the exception that's being thrown.

There is also some "Papiea encountered unexpected error" wrapping which shows up in the log; I'm not sure where this is coming from (or whether it's relevant yet).

This gets worse when crossing API boundaries—I don't understand how/why this is happening yet, but I have seen one case in which a procedure was calling a custom constructor, which threw an exception, and somehow by the time the error made it back to the client (who called the procedure), the error had been lost entirely and was simply being reported as `Procedure invocation failed.` with a sub-error `Error while making request` and no further details.

### Expected/Actual Behavior

I expect the exception raised in the custom constructor to be propagated back through the procedure (as an API-client exception, which then bubbles up to the SDK/context) and to the caller without information being lost.

### Steps to Reproduce
The specific steps are highly provider-dependent; I can share logs offline which show the entire path of the exception in question.

### Impact
Wasted time digging through logs to find more complete error information. This gets especially bad if the people doing the testing don't have complete access to dig around in the internal logs for Papiea/the provider themselves (and need to involve someone from another team to get a complete picture of what happened).

### Additional Details

I think this will require some design/design thinking to understand and clearly define how errors flow through the system, especially when multiple providers and/or nested calls through Papiea are involved. Especially because we need to be careful, once we start using Papiea in production, about what error information we return to the caller.

But, even without a complete design, if we can take some immediate steps to avoid re-wrapping exceptions and make sure they are propagated with complete information, that would be really helpful.

<hr>

### LINKED ISSUE #658: Audit Papiea for improper error handling
There are a few places I've run across in Papiea where we have exception handling that looks something like this:

```ts
try {
    // do something
    return true; // success
} catch (e) {
    return false; // failed
}
```

Nitesh and I ran across a similar example in `diff_resolver.ts` earlier today:

```ts
        try {
            const updated_watchlist = await this.watchlistDb.get_watchlist()
            this.watchlist.update(updated_watchlist)
        } catch (e) {
            return
        }
```

We should never, ever write code that looks like this, because unexpected/unanticipated errors will get swallowed and nobody will be able to debug what's going on.

All our catch blocks should look something like:

```ts
try {
    ...
} catch (e) {
    // First, never assume ANYTHING about `e`.
    // It's perfectly valid to `throw undefined`!

    // For errors that get swallowed/handled specially, we must look
    // for some specific error:
    if (e instanceof ...) { ... }
    if (e?.response?.status == 400) { /* do whatever */ }

    // After checking for specific errors, there are two acceptable
    // ways to end a catch block:

    // Option 1: Re-throw it:
    throw e;

    // Option 2 (in rare cases only): Swallow and log it:
    if (typeof e === 'object') {
        console.log(e.toString());
        console.log(e.stack); // Always log stack if possible
    } else {
        console.log(e.toString());
    }
}
```

We need to audit the codebase, find places where we are swallowing errors or otherwise catching things in a way that doesn't follow the above pattern, and fix them.

nitesh-idnani1:
Closes #633
Closes #658

### DIFF
--- a/papiea-client/src/errors/errors.ts
+++ b/papiea-client/src/errors/errors.ts
@@ -1,7 +1,9 @@
-export class PapieaErrorImpl extends Error {
-    original_error: Error
+import { PapieaError } from "papiea-core"
 
-    constructor(message: string, e: Error) {
+export class PapieaErrorImpl extends Error {
+    original_error: any
+
+    constructor(message: string, e: any) {
         super(message)
         this.original_error = e
     }
@@ -10,45 +12,79 @@ export class PapieaErrorImpl extends Error {
 
 // Base exception received from papiea
 export class PapieaException extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.PapieaException
+    }
 }
 
 // Spec with this version already exists
 export class ConflictingEntityError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.ConflictingEntity
+    }
 }
 
 // Entity not found on papiea
 export class EntityNotFoundError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.EntityNotFound
+    }
 }
 
 // Token provided doesn't have access rights for the operation
 export class PermissionDeniedError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.PermissionDenied
+    }
 }
 
 // Error in procedure handler
 export class ProcedureInvocationError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.ProcedureInvocation
+    }
 }
 
 // No auth token provided in Authorization header
 export class UnauthorizedError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.Unauthorized
+    }
 }
 
 // Entity spec/status didn't pass validation
 export class ValidationError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.Validation
+    }
 }
 
 // Wrong data on the request side
 export class BadRequestError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.BadRequest
+    }
 }
 
+// Error in on_create/on_delete/intent handlers
+export class OnActionError extends PapieaErrorImpl {
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.OnActionError
+    }
+}
 // Something went wrong on papiea
 export class PapieaServerError extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.ServerError
+    }
 }
--- a/papiea-client/src/errors/errors.ts
+++ b/papiea-client/src/errors/errors.ts
@@ -1,7 +1,7 @@
 import { PapieaError } from "papiea-core"
 
 export class PapieaErrorImpl extends Error {
-    original_error: any
+    original_error: Error
 
     constructor(message: string, e: any) {
         super(message)
--- a/papiea-engine/__tests__/database_tests/database_mongo.test.ts
+++ b/papiea-engine/__tests__/database_tests/database_mongo.test.ts
@@ -106,7 +106,7 @@ describe("MongoDb tests", () => {
             await specDb.update_spec(entity_metadata, spec);
         } catch (err) {
             expect(err).toBeInstanceOf(ConflictingEntityError);
-            expect(err.entity_info.additional_info.existing_spec_version).toEqual("1");
+            expect(err.entity_info.additional_info.existing_spec_version).toEqual("2");
         }
     });
 
--- a/papiea-engine/__tests__/database_tests/database_mongo.test.ts
+++ b/papiea-engine/__tests__/database_tests/database_mongo.test.ts
@@ -106,7 +106,7 @@ describe("MongoDb tests", () => {
             await specDb.update_spec(entity_metadata, spec);
         } catch (err) {
             expect(err).toBeInstanceOf(ConflictingEntityError);
-            expect(err.existing_metadata.spec_version).toEqual(2);
+            expect(err.entity_info.additional_info.existing_spec_version).toEqual("1");
         }
     });
 
--- a/papiea-engine/__tests__/entity_tests/entity_api.test.ts
+++ b/papiea-engine/__tests__/entity_tests/entity_api.test.ts
@@ -60,7 +60,7 @@ describe("Entity API tests", () => {
     });
 
     test("Create entity with malformed spec should fail", async () => {
-        expect.assertions(2);
+        expect.assertions(1);
         try {
             await entityApi.post(`/${providerPrefix}/${providerVersion}/${kind_name}`, {
                 spec: {
@@ -71,7 +71,6 @@ describe("Entity API tests", () => {
         } catch (err) {
             const res = err.response;
             expect(res.status).toEqual(400);
-            expect(res.data.error.errors.length).toEqual(1);
         }
     });
 
@@ -351,7 +350,7 @@ describe("Entity API tests", () => {
     });
 
     test("Update entity with malformed spec should fail", async () => {
-        expect.assertions(3);
+        expect.assertions(2);
         try {
             const { data: { metadata, spec } } = await entityApi.post(`/${providerPrefix}/${providerVersion}/${kind_name}`, {
                 spec: {
@@ -372,7 +371,6 @@ describe("Entity API tests", () => {
         } catch (err) {
             const res = err.response;
             expect(res.status).toEqual(400);
-            expect(res.data.error.errors.length).toEqual(1);
         }
     });
 
@@ -437,7 +435,7 @@ describe("Entity API tests", () => {
             });
         } catch (e) {
             expect(e.response.status).toEqual(409)
-            expect(e.response.data.error.message).toEqual(`Conflicting Entity: ${entity_uuid}. Existing entity has version ${1}`)
+            expect(e.response.data.error.error_details.message).toEqual(`Entity with UUID ${entity_uuid} of kind: ${providerPrefix}/${providerVersion}/${kind_name} already exists.`)
         }
     });
 
@@ -519,7 +517,7 @@ describe("Entity API tests", () => {
             });
         } catch (e) {
             expect(e.response.status).toEqual(409)
-            expect(e.response.data.error.message).toEqual(`Deleted entity with this uuid and spec version exists: uuid - ${id}, maximum current spec version - 2`)
+            expect(e.response.data.error.error_details.message).toEqual(`Deleted entity with UUID ${id} of kind: ${providerPrefix}/${providerVersion}/${kind_name} already exists with this spec version.`)
         }
     });
 
--- a/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
+++ b/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
@@ -153,9 +153,9 @@ describe("Procedures tests", () => {
             expect(res.status).toEqual(500);
             expect(res.data.error.message).toEqual("Procedure Invocation Failed")
             expect(res.data.error.code).toEqual(500)
-            const error_messages = JSON.parse(res.data.error.error_details.cause.cause.message)
-            expect(error_messages[0]).toContain(`Received procedure input is missing required field: x`);
-            expect(error_messages[1]).toContain(`Received procedure input is missing required field: y`);
+            const errors = res.data.error.error_details.cause.cause.errors
+            expect(errors[0].message).toContain(`Received procedure input is missing required field: x`);
+            expect(errors[1].message).toContain(`Received procedure input is missing required field: y`);
         }
     });
 
--- a/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
+++ b/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
@@ -103,7 +103,6 @@ describe("Procedures tests", () => {
         } catch (err) {
             const res = err.response;
             expect(res.status).toEqual(400);
-            expect(res.data.error.errors.length).toEqual(1);
             return;
         }
     });
@@ -120,10 +119,10 @@ describe("Procedures tests", () => {
         } catch (err) {
             const res = err.response;
             expect(res.status).toEqual(400);
-            expect(res.data.error.errors.length).toEqual(1);
-            expect(res.data.error.errors[0].message).toContain(`Procedure was expecting non-empty object, received null/empty object`);
+            expect(res.data.error.error_details.message).toContain(`Procedure was expecting non-empty object, received null/empty object`);
         }
     });
+
     test("Procedure result validation", async () => {
         expect.hasAssertions();
         const server = http.createServer((req, res) => {
@@ -152,11 +151,10 @@ describe("Procedures tests", () => {
         } catch (err) {
             const res = err.response;
             expect(res.status).toEqual(500);
-            expect(res.data.error.errors.length).toEqual(2);
-            expect(res.data.error.message).toEqual("Procedure invocation failed.")
+            expect(res.data.error.message).toEqual("Procedure Invocation Failed")
             expect(res.data.error.code).toEqual(500)
-            expect(res.data.error.errors[0].message).toContain(`Received procedure input is missing required field: x`);
-            expect(res.data.error.errors[1].message).toContain(`Received procedure input is missing required field: y`);
+            expect(res.data.error.error_details.message).toContain(`Received procedure input is missing required field: x`);
+            expect(res.data.error.error_details.message).toContain(`Received procedure input is missing required field: y`);
         }
     });
 
--- a/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
+++ b/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
@@ -153,8 +153,9 @@ describe("Procedures tests", () => {
             expect(res.status).toEqual(500);
             expect(res.data.error.message).toEqual("Procedure Invocation Failed")
             expect(res.data.error.code).toEqual(500)
-            expect(res.data.error.error_details.message).toContain(`Received procedure input is missing required field: x`);
-            expect(res.data.error.error_details.message).toContain(`Received procedure input is missing required field: y`);
+            const error_messages = JSON.parse(res.data.error.error_details.cause.cause.message)
+            expect(error_messages[0]).toContain(`Received procedure input is missing required field: x`);
+            expect(error_messages[1]).toContain(`Received procedure input is missing required field: y`);
         }
     });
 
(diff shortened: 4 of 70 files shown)


nitesh-idnani1:
I'm facing some issues with adding special characters like "\n" or "\t" in JSON strings, they are getting parsed as raw characters on the user end and making the error messages look a bit messy. Any idea on how to fix this?
Example:
https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/src/errors/papiea_exception.ts#L15

nitesh-idnani1:
As evident this is a large PR with changes in a lot of files, if you want to review the major changes related to design you can look at the following files:
https://github.com/nutanix/papiea/blob/nitesh/fix-error-details-swallowing/papiea-core/src/core.ts
https://github.com/nutanix/papiea/blob/nitesh/fix-error-details-swallowing/papiea-engine/src/databases/utils/errors.ts
https://github.com/nutanix/papiea/tree/nitesh/fix-error-details-swallowing/papiea-engine/src/errors
https://github.com/nutanix/papiea/blob/nitesh/fix-error-details-swallowing/papiea-engine/src/utils/utils.ts
https://github.com/nutanix/papiea/blob/nitesh/fix-error-details-swallowing/papiea-sdk/python/papiea/python_sdk.py
https://github.com/nutanix/papiea/blob/nitesh/fix-error-details-swallowing/papiea-sdk/python/papiea/python_sdk_exceptions.py
https://github.com/nutanix/papiea/blob/nitesh/fix-error-details-swallowing/papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
https://github.com/nutanix/papiea/blob/nitesh/fix-error-details-swallowing/papiea-sdk/typescript/src/provider_sdk/typescript_sdk_exceptions.ts

All the other files include tests modifications/throw statements modification.

papiea-client/src/errors/errors.ts
```diff
@@ -1,7 +1,9 @@
+import { PapieaError } from "papiea-core"
+
 export class PapieaErrorImpl extends Error {
-    original_error: Error
+    original_error: any
```
Emulebest, line 4: Why is this type `any`?

papiea-client/src/errors/errors.ts
```diff
@@ -10,45 +12,79 @@ export class PapieaErrorImpl extends Error {
 
 // Base exception received from papiea
 export class PapieaException extends PapieaErrorImpl {
-
+    constructor(message: string, e: any) {
+        super(message, e)
+        this.name = PapieaError.PapieaException
```
Emulebest, line 17: I think this is more of a `this.type` than name

papiea-engine/__tests__/database_tests/database_mongo.test.ts
```diff
@@ -106,7 +106,7 @@ describe("MongoDb tests", () => {
             await specDb.update_spec(entity_metadata, spec);
         } catch (err) {
             expect(err).toBeInstanceOf(ConflictingEntityError);
-            expect(err.existing_metadata.spec_version).toEqual(2);
+            expect(err.entity_info.additional_info.existing_spec_version).toEqual("1");
```
Emulebest, line 109: Why is this now `toEqual("1")` rather than `toEqual(2)`. Seems like a bug that leads to this test being broken

papiea-engine/__tests__/entity_tests/entity_api.test.ts
```diff
@@ -519,7 +517,7 @@ describe("Entity API tests", () => {
             });
         } catch (e) {
             expect(e.response.status).toEqual(409)
-            expect(e.response.data.error.message).toEqual(`Deleted entity with this uuid and spec version exists: uuid - ${id}, maximum current spec version - 2`)
+            expect(e.response.data.error.error_details.message).toEqual(`Deleted entity with UUID ${id} of kind: ${providerPrefix}/${providerVersion}/${kind_name} already exists with this spec version.`)
```
Emulebest, line 520: Could we substitute the `of kind:` with `of: `. Applies to all of the below

papiea-engine/src/entity/entity_routes.ts
```diff
@@ -218,7 +218,7 @@ function queryToString(q?: ExpressQueryParam, fieldName?: string): string | unde
         case 'string': return q;
         case 'undefined': return undefined;
         default:
-            throw new BadRequestError(`Query parameter: ${fieldName} has invalid type: ${typeof q}`);
+            throw new BadRequestError({ message: `Query parameter: ${fieldName} has invalid type: ${typeof q}, expected (string, undefined).` });
```
Emulebest, line 221: Expected undefined?

papiea-engine/src/errors/papiea_error_impl.ts
```diff
@@ -10,108 +10,142 @@ import { PermissionDeniedError, UnauthorizedError } from "./permission_error";
 import { BadRequestError } from "./bad_request_error";
 import { PapieaError } from "papiea-core";
 import {Logger} from "papiea-backend-utils"
-import {PapieaExceptionContextImpl, PapieaException} from "../errors/papiea_exception"
+import { PapieaException } from "../errors/papiea_exception"
 import { OnActionError } from "./on_action_error";
 
 export class PapieaErrorResponseImpl implements PapieaResponse {
```
Emulebest, line 16: I think at this point this could be called `PapieaErrorResponseProcessor`

papiea-engine/src/intentful_core/intentful_strategies/intentful_strategy_interface.ts
```diff
@@ -65,11 +65,11 @@ export abstract class IntentfulStrategy {
...
                 }
             }
         } else {
-            throw OnActionError.create(`Could not delete the entity since kind ${entity.metadata?.provider_prefix}/${entity.metadata?.provider_version}/${entity.metadata?.kind} is not registered`, procedure_name, { provider_prefix: entity.metadata?.provider_prefix, provider_version: entity.metadata?.provider_version, kind_name: entity.metadata?.kind, additional_info: { "entity_uuid": entity.metadata?.uuid ?? '', "procedure_name": procedure_name }})
+            throw new OnActionError({ message: `Could not delete the entity since kind: ${entity.metadata?.provider_prefix}/${entity.metadata?.provider_version}/${entity.metadata?.kind} is not registered.`, entity_info: { provider_prefix: entity.metadata?.provider_prefix, provider_version: entity.metadata?.provider_version, kind_name: entity.metadata?.kind, additional_info: { "entity_uuid": entity.metadata?.uuid ?? '', "procedure_name": procedure_name }}})
```
Emulebest, line 72: since kind?

papiea-engine/src/utils/utils.ts
```diff
@@ -174,12 +180,63 @@ export function getVersionVerifier(enginePapieaVersion: string) {
...
             if (semver.diff(headersPapieaVersion, enginePapieaVersion) === 'major') {
-                throw new BadRequestError(`Received incompatible papiea version: ${headersPapieaVersion}, expected version compatible with ${headersPapieaVersion}`)
+                throw new BadRequestError({ message: `Received incompatible papiea version: ${headersPapieaVersion}, expected version compatible with ${headersPapieaVersion}` })
             }
         }
         next();
     }
 }
+
+export function convertEntityInfoToExceptionContext(entity_info: any): PapieaExceptionContext {
```
Emulebest, line 193: Can we move this to be a static method of `PapieaErrorResponseImpl`?

papiea-engine/src/utils/utils.ts
```diff
@@ -174,12 +180,63 @@ export function getVersionVerifier(enginePapieaVersion: string) {
...
+
+    for (let field in entity_info) {
+        if (field !== "provider_prefix" && field !== "provider_version" && field !== "kind_name") {
+            additional_info[field] = entity_info[field]
+        }
+    }
+
+    return { provider_prefix, provider_version, kind_name, additional_info }
+}
+export function convertAxiosErrorToEngineError(error: any): Error {
```
Emulebest, line 207: Can we move this to be a static method of `PapieaErrorResponseImpl`?

Emulebest (changes requested):
Overall the PR looks really good. (Note the `for kind:` change request). One thing I am concerned about is how do we handle multiple errors and/or nested errors. Please create a test that:
1) Tries to create a kind with 2 or more incorrect schema parts
2) Catch both errors in the engine via validator
3) Propagate and check both error messages in a single papiea-engine error response

nitesh-idnani1:
> Overall the PR looks really good. (Note the `for kind:` change request). One thing I am concerned about is how do we handle multiple errors and/or nested errors. Please create a test that:
> 
> 1. Tries to create a kind with 2 or more incorrect schema parts
> 2. Catch both errors in the engine via validator
> 3. Propagate and check both error messages in a single papiea-engine error response

Already added assert checks for nested and multiple errors:
Nested error - https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts#L2350
Multiple error - https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts#L156

papiea-client/src/errors/errors.ts
nitesh-idnani1: Yeah agreed, I'm just re-using the type field extended by the Javascript Error class and taken care to print it as type whenever we include it in the response.
Reference: https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/src/errors/papiea_exception.ts#L33
https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/src/errors/papiea_error_impl.ts#L80

papiea-engine/__tests__/entity_tests/entity_api.test.ts
nitesh-idnani1: Added that to clarify what we're printing, since we have just provider information as some places and kind information at others.

papiea-engine/src/utils/utils.ts
nitesh-idnani1: This method is being used by the Procedure Invocation error to process the axios error and convert it to a suitable Error object. Since, the Procedure Invocation error does not have any context of the error response I decided to make this and "convertEntityInfoToExceptionContext" as util helpers.
Usage Reference:
https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/src/errors/procedure_invocation_error.ts#L24

Emulebest: Still I don't see a reason for this to become `PapieaErrorResponseImpl.convertAxiosErrorToEngineError(err.response?.data.error)`. This function is too specific to be a global util IMO

Emulebest:
> papiea/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts

It seems like it is spliced into one single string, I don't like that. Let's go over that on the call

Emulebest:
> > Overall the PR looks really good. (Note the `for kind:` change request). One thing I am concerned about is how do we handle multiple errors and/or nested errors. Please create a test that:
> > 
> > 1. Tries to create a kind with 2 or more incorrect schema parts
> > 2. Catch both errors in the engine via validator
> > 3. Propagate and check both error messages in a single papiea-engine error response
> 
> Already added assert checks for nested and multiple errors:
> Nested error -
> 
> https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts#L2350
> 
> 
> Multiple error -
> https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts#L156

It seems like it is spliced into one single string, I don't like that. Let's go over that on the call

nitesh-idnani1:
Analyzed the requirements for validation error type, there is only one module wherein we throw multiple errors within the validation error "cause".
So in order to avoid complicating the error design, I modified the validator module to generate multiple error in an list format which the user can parse (JSON.parse) and then go through each error message as an individual item.
Also, I realized the stack trace for each error type is not that relevant since the ValidationError already encapsulates the trace upto the error cause, and since we hide the trace from user I felt it's too much redundancy to keep trace for each individual error.

nitesh-idnani1:
> > > Overall the PR looks really good. (Note the `for kind:` change request). One thing I am concerned about is how do we handle multiple errors and/or nested errors. Please create a test that:
> > > 
> > > 1. Tries to create a kind with 2 or more incorrect schema parts
> > > 2. Catch both errors in the engine via validator
> > > 3. Propagate and check both error messages in a single papiea-engine error response
> > 
> > 
> > Already added assert checks for nested and multiple errors:
> > Nested error -
> > https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts#L2350
> > 
> > Multiple error -
> > https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts#L156
> 
> It seems like it is spliced into one single string, I don't like that. Let's go over that on the call

Analyzed the requirements for validation error type, there is only one module wherein we throw multiple errors within the validation error "cause".
So in order to avoid complicating the error design, I modified the validator module to generate multiple error in an list format which the user can parse (JSON.parse) and then go through each error message as an individual item.
Also, I realized the stack trace for each error type is not that relevant since the ValidationError already encapsulates the trace upto the error cause, and since we hide the trace from user I felt it's too much redundancy to keep trace for each individual error.

Emulebest:
> > > > Overall the PR looks really good. (Note the `for kind:` change request). One thing I am concerned about is how do we handle multiple errors and/or nested errors. Please create a test that:
> > > > 
> > > > 1. Tries to create a kind with 2 or more incorrect schema parts
> > > > 2. Catch both errors in the engine via validator
> > > > 3. Propagate and check both error messages in a single papiea-engine error response
> > > 
> > > 
> > > Already added assert checks for nested and multiple errors:
> > > Nested error -
> > > https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts#L2350
> > > 
> > > Multiple error -
> > > https://github.com/nutanix/papiea/blob/061403b097330fb5a814111c46f92d2d403f6a10/papiea-engine/__tests__/procedure_tests/procedures_api.test.ts#L156
> > 
> > 
> > It seems like it is spliced into one single string, I don't like that. Let's go over that on the call
> 
> Analyzed the requirements for validation error type, there is only one module wherein we throw multiple errors within the validation error "cause".
> So in order to avoid complicating the error design, I modified the validator module to generate multiple error in an list format which the user can parse (JSON.parse) and then go through each error message as an individual item.
> Also, I realized the stack trace for each error type is not that relevant since the ValidationError already encapsulates the trace upto the error cause, and since we hide the trace from user I felt it's too much redundancy to keep trace for each individual error.

If the error design has some cases that requires complication then let's do this. Asking users to sometimes do `JSON.parse` is a bad thing IMO. For the error list you can take inspiration from the current design. It has the `error.errors` array field, if it has a single error, then it is a single item list, otherwise it has multiple errors like validation failures.

shlomi-nx (commented):
Looks good so far, see comments below

papiea-engine/__tests__/procedure_tests/procedures_api.test.ts
```diff
@@ -152,11 +151,11 @@ describe("Procedures tests", () => {
         } catch (err) {
             const res = err.response;
             expect(res.status).toEqual(500);
-            expect(res.data.error.errors.length).toEqual(2);
-            expect(res.data.error.message).toEqual("Procedure invocation failed.")
+            expect(res.data.error.message).toEqual("Procedure Invocation Failed")
             expect(res.data.error.code).toEqual(500)
-            expect(res.data.error.errors[0].message).toContain(`Received procedure input is missing required field: x`);
-            expect(res.data.error.errors[1].message).toContain(`Received procedure input is missing required field: y`);
+            const error_messages = JSON.parse(res.data.error.error_details.cause.cause.message)
```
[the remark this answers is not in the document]
nitesh-idnani1: Good observation but we should segregate this to a separate PR (adding helper modules for axios)

papiea-engine/src/intentful_core/entity_creation_strategies/constructor_entity_creation_strategy.ts
```diff
@@ -53,14 +53,14 @@ export class ConstructorEntityCreationStrategy extends EntityCreationStrategy {
...
             }
         }
         const spec_status_equal = deepEqual(entity.spec, entity.status)
         if (!spec_status_equal && this.kind.intentful_behaviour === IntentfulBehaviour.SpecOnly) {
-            throw OnActionError.create(`Spec-only entity constructor returned of kind ${this.provider.prefix}/${this.provider.version}/${this.kind.name} spec not matching status`, `__${this.kind.name}_create`, { provider_prefix: this.provider.prefix, provider_version: this.provider.version, kind_name: this.kind.name, additional_info: { "entity_uuid": entity.metadata.uuid, "procedure_name": `__${this.kind.name}_create` }})
+            throw new OnActionError({ message: `Spec-only entity constructor returned spec not matching status for entity of kind: ${this.provider.prefix}/${this.provider.version}/${this.kind.name}.`, entity_info: { provider_prefix: this.provider.prefix, provider_version: this.provider.version, kind_name: this.kind.name, additional_info: { "entity_uuid": entity.metadata.uuid, "procedure_name": `__${this.kind.name}_create` }}})
         }
         const span = spanOperation(`save_entity_db`,
                                    ctx.tracing_ctx)
```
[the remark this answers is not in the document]
nitesh-idnani1: Yes, I'm seeing we need a better representation for the Jaeger span. I'm working on this as part of the "Add log tags" (#575) issue. Let's discuss about this tomorrow.

papiea-engine/src/intentful_engine/intent_resolver.ts
```diff
@@ -181,7 +181,7 @@ export class IntentResolver {
                     const [, status] = await this.statusDb.get_status({...entry_ref.provider_reference, ...entry_ref.entity_reference})
                     this.onChange({ metadata, spec, status })
                 } catch (e) {
-
+                    console.debug(`Failed to process onChange in update active watcher status for entity with uuid: ${entry_ref.entity_reference.uuid} and kind: ${entry_ref.entity_reference.kind} for provider with prefix: ${entry_ref.provider_reference.provider_prefix} and version: ${entry_ref.provider_reference.provider_version} due to error: ${e}`)
```
Emulebest, line 184: Why is this a `console.debug` and not a `logger.debug`?

papiea-engine/src/auth/s2s.ts
```diff
@@ -18,6 +18,7 @@ export class S2SKeyUserAuthInfoExtractor implements UserAuthInfoExtractor {
             user_info.authorization = 'Bearer ' + s2skey.key;
             return user_info;
         } catch (e) {
+            console.debug(`Failed to get user auth info for provider ${provider_prefix}/${provider_version} due to error: ${e}`)
```
Emulebest, line 21: Why is this a `console.debug` and not a `logger.debug`?

Emulebest (changes requested):
A few comments

Emulebest:
Please also update this to latest master

Emulebest (approved):

Emulebest:
@nitesh-idnani1 please resolve merge conflicts and merge it in, thanks

PR #645, opened 2021-03-03
Remove undefined and null fields from the differ input

### LINKED ISSUE #644: Undefined fields are being cast to null fields which throws off the differ
We found a serious problem on the boundary between javascript and clojure. Apparently, on some occasions,  the differ is receiving spec and status that contains undefined fields. We couldn't see that its the case when printing them, because js "conveniently" hides them, but when transforming the data-structures to clojures using `js->clj`, all undefined fields are turned to null fields. This had the effect of showing us diffs where there were no diffs.

nitesh-idnani1:
Closes #644
- Added helper function to remove undefined fields from status and spec passed to differ input.
- Added tests to verify the changes.

### DIFF
--- a/papiea-core/src/core.ts
+++ b/papiea-core/src/core.ts
@@ -141,14 +141,11 @@ export interface DiffContent {
 export interface Differ {
 
     // Get the next diff from an entity based on the 
-    diffs(kind: Kind, spec: Spec, status: Status): Generator<Diff, any, undefined>;
+    diffs(kind: Kind, spec: Spec, status: Status, logger?: any): Generator<Diff, any, undefined>;
 
     // We could also get the entire list of diffs, ordered by the
     // original dependency tree
-    all_diffs(kind: Kind, spec: Spec, status: Status): Diff[];
-
-    // Removes status-only fields from the entity status using the schema
-    remove_status_only_fields(schema: any, status: Status): Status;
+    all_diffs(kind: Kind, spec: Spec, status: Status, logger?: any): Diff[];
 
     // Get current value by path specified in diff fields
     get_diff_path_value(diff: DiffContent, spec: Spec): any
--- a/papiea-engine/src/intentful_core/differ_impl.ts
+++ b/papiea-engine/src/intentful_core/differ_impl.ts
@@ -1,12 +1,13 @@
 import { Spec, Status, Kind, Differ, Diff, DiffContent } from "papiea-core"
+import { Logger } from "papiea-backend-utils";
 import { SFSCompiler } from "./sfs_compiler"
 
 export class BasicDiffer implements Differ {
     // Get the diff iterator from an entity based on the
-    public* diffs(kind: Kind, spec: Spec, status: Status): Generator<Diff, any, undefined> {
+    public* diffs(kind: Kind, spec: Spec, status: Status, logger?: Logger): Generator<Diff, any, undefined> {
         for (let sig of kind.intentful_signatures) {
             const compiled_signature = SFSCompiler.try_compile_sfs(sig.signature, kind.name)
-            const result = SFSCompiler.run_sfs(compiled_signature, spec, status)
+            const result = SFSCompiler.run_sfs(compiled_signature, spec, status, kind.kind_structure[kind.name], kind.name,logger)
             if (result != null && result.length > 0) {
                 yield {
                     kind: kind.name,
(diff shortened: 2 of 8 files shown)


papiea-engine/src/intentful_engine/diff_resolver.ts
```diff
@@ -109,8 +109,10 @@ export class DiffResolver {
                 diff_status = JSON.parse(JSON.stringify(status));
             }
             diff_status = this.differ.remove_status_only_fields(kind.kind_structure[kind.name], diff_status);
+            const diff_spec = this.differ.remove_undefined_or_null_values(spec, kind.name)
+            diff_status = this.differ.remove_undefined_or_null_values(diff_status, kind.name, true)
             return {
-                diffs: this.differ.all_diffs(kind, spec, diff_status),
+                diffs: this.differ.all_diffs(kind, diff_spec, diff_status),
                 metadata, provider, kind, spec, status,
             };
```
shlomi-nx, line 117: I don't think that this is necessarily the best place to sanitize, see my second post on the issue https://github.com/nutanix/papiea/issues/644#issuecomment-789346699

Instead this should get sanitized just before invoking the clojure code, at https://github.com/nutanix/papiea/blob/dfb71930b9192723b7a1d4ddc01e6199cd23924b/papiea-engine/src/intentful_core/sfs_compiler.ts#L30-L32

shlomi-nx (changes requested):
Looks good but needs a few changes, see below:

papiea-engine/src/intentful_core/differ_impl.ts
```diff
@@ -63,6 +63,30 @@ export class BasicDiffer implements Differ {
...
+                v.forEach(v_item => {
+                    newObj[k].push(this.remove_undefined_or_null_values(v_item, kind_name, isStatus, field_name + "/" + k))
+                });
+            } else if (v === Object(v)) {
+                newObj[k] = this.remove_undefined_or_null_values(v, kind_name, isStatus, field_name + "/" + k);
+            } else if (v != null && v !== undefined) {
+                newObj[k] = entity[k];
+            } else {
+                if (isStatus) {
+                    console.debug(`Removing undefined/null field: ${field_name}/${k} from the entity status for kind ${kind_name}.`);
```
Emulebest, line 81: Please use a logger here instead of `console`

papiea-core/src/core.ts
```diff
@@ -150,6 +150,9 @@ export interface Differ {
     // Removes status-only fields from the entity status using the schema
     remove_status_only_fields(schema: any, status: Status): Status;
 
+    // Remove all the null and undefined values from the entity
+    remove_undefined_or_null_values(entity: any, kind_name?: string, isStatus?: boolean, field_name?: string): any
+
```
Emulebest, line 155: Please expose a function `sanitize` in the interface and hide `remove_status_only_fields` and `remove_undefined_or_null_values` behind implementation

papiea-engine/src/intentful_engine/diff_resolver.ts
Emulebest: Agree, diff_resolver isn't a good place to do this in

Emulebest (changes requested):

papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -1289,6 +1289,62 @@ describe("Provider Sdk tests", () => {
...
+                }
+            });
+
+            const updatedEntity: any = await axios.get(`${sdk.entity_url}/${sdk.provider.prefix}/${sdk.provider.version}/${kind_name}/${metadata.uuid}`);
+            await axios.put(`${sdk.entity_url}/${sdk.provider.prefix}/${sdk.provider.version}/${kind_name}/${metadata.uuid}`, {
+                spec: {
+                    test: null
+                },
+                metadata: metadata
+            });
```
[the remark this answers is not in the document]
nitesh-idnani1: Yeah I tried with undefined but axios has it's own handler for undefined fields which is same as what we do (removing the fields), so I wasn't able to write a test for undefined. But I did try out a value check within the engine itself and it works as expected for undefined.

papiea-engine/src/intentful_core/sfs_compiler.ts
```diff
@@ -11,6 +12,74 @@ const sfs_compiler = (sfs_signature: string) => papi_clj.core.compile_sfs(sfs_si
...
+}
+
+function cleanup_spec_for_sfs_run(spec: Spec, kind_name: string, logger?: Logger): Spec {
+    let diff_spec: Spec = JSON.parse(JSON.stringify(spec));
+    diff_spec = remove_undefined_or_null_values(diff_spec, logger, kind_name, "spec")
+    return diff_spec
+}
+
+function cleanup_status_for_sfs_run(status: Status, schema: any, kind_name: string, logger?: Logger): Status {
+    let diff_status: Status = JSON.parse(JSON.stringify(status));
```
shlomi-nx, line 77: I am not sure we need this line, as it does the same as `remove_undefined_or_null_values`, only more expensive

shlomi-nx (commented):
Looks great! Just one redundant line that I think we should remove (in both places) and we should be ready to go!

papiea-engine/src/intentful_core/sfs_compiler.ts
nitesh-idnani1: I added this to ensure we copy spec/status by value, otherwise the original spec/status gets modified after the cleanup.

shlomi-nx: hmm, kind of like a deep-clone? darn mutable languages.. 

Can the sanitizer itself simply return a cloned version? I mean it anyways traverses the entire tree. It's just seems pretty wasteful at this point..

shlomi-nx (approved):
Looks great, lets merge once it passes CI

PR #648, opened 2021-03-04
Remove nullable fields

### LINKED ISSUE #611: Make sure nullable fields are handled properly or excluded
As part of #596, I realized there may be a conflict in how Papiea is handling schemata vs. its `update_status` behavior. Specifically, if a field is `nullable` per [the Swagger spec](https://swagger.io/specification/#schema-nullable), but not `required`, it is possible the field could be absent entirely, have an explicit value of `null`, or have an actual value.

There is no way to correctly set such a field to `null` with `update_status()` right now, because `update_status()` assumes that explicitly setting a field to `null` should remove it entirely.

We should either figure out a way to differentiate field removal vs. explicitly assigning `null` in the `update_status()` API, or we should disallow `nullable` fields entirely.

Emulebest:
resolves #611

### DIFF
--- a/papiea-engine/__tests__/provider_tests/provider_api.test.ts
+++ b/papiea-engine/__tests__/provider_tests/provider_api.test.ts
@@ -256,56 +256,15 @@ describe("Provider API tests", () => {
         }
     });
 
-    test("Partial Update status with undefined values should be meaningful", async () => {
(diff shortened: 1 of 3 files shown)


shlomi-nx (approved):
Looks good

PR #649, opened 2021-03-05
Modified logic for the sanitizer functions in sfs compiler

### LINKED ISSUE #644: Undefined fields are being cast to null fields which throws off the differ

nitesh-idnani1:
Closes #644
- Modified logic for remove undefined/null values to avoid corner cases.


Emulebest (approved):
@nitesh-idnani1 Can we transfer some test ideas from @joshua-berry-ntnx for this not to happen again?

nitesh-idnani1:
> @nitesh-idnani1 Can we transfer some test ideas from @joshua-berry-ntnx for this not to happen again?

Yes, good idea. I'll ask him for inputs

PR #664, opened 2021-03-18
Added debug log messages with related tags in papiea engine

### LINKED ISSUE #575: Add log tags and dynamic log levels

nitesh-idnani1:
Closes #575
- Added log messages to indicate the execution of function within the engine.
- Added suitable log tags to identify the engine components.


shlomi-nx (commented):
Looks very good, a couple of suggestions:
1. Perhaps we should make the tags an enum, to easily communicate the different tags we support
2. The fields says `tags` in plural, but only allows for a single tag. Either make it a list of tags, or make the name singular

nitesh-idnani1:
> Looks very good, a couple of suggestions:
> 
> 1. Perhaps we should make the tags an enum, to easily communicate the different tags we support
> 2. The fields says `tags` in plural, but only allows for a single tag. Either make it a list of tags, or make the name singular

Fixed the comments, thanks @shlomi-nx.

Emulebest:
@nitesh-idnani1 please update this to the latest master, a lot of changes I see pending

joshua-berry-ntnx (changes requested):
TBH a lot of this seems excessive to me, and I think will ultimately have the opposite of its intended effect, since there is both too much and not enough info:

* "Too much" because we are logging at both entry and (most) exits in most major functions, when simply logging the outcome of a particular function is often sufficient, or when logging logging that function doesn't actually provide much additional value (e.g. because there are other logs in other places that already tell you what you need to know, or because what the function does is pretty trivial). 
* "Not enough" because often we are logging little more than the function name—not its arguments, return value, side effects, etc., and so there is not enough context/info about the specific scenario to tell what happened or what should have happened.

So you will wind up with a bunch of logs that look like:
```
BEGIN foo in bar {tags: [whatever]}
BEGIN subfoo in baz {tags: [whatever]}
END subfoo in baz {tags: [whatever]}
END foo in bar {tags: [whatever]}
```

...but no idea why foo returned `undefined` when it should have returned `47`, or why the validator threw some exception when it should have succeeded.

IMO the key to writing good log messages (and error messages) is to ask, "What is this log telling me that I cannot get from reading the code or from somewhere else (e.g. poking around in the DB/UI/etc.)? What context does the reader need to have to understand what's going on here?"

For example, I don't need logs to tell me that an entity was fetched from the DB. I can tell that from the code. But I DO need logs to tell me things like, "fetched entity X which had owner Y, and owner Y doesn't match owner A for key B, so we rejected the request as unauthorized".

The places where logs are most helpful are typically places like these:
* The code is making some consequential decision, and we need to capture how and why the decision was made (and of course the decision/outcome itself)
* Inputs, outputs and intermediate steps of more complex/consequential processes, so we can see how an output was reached
* Anything "unexpected" (for a wide definition of "unexpected")
* Failures (including why/how the failures happened, and any causes/factors that led up to the failure, to the extent we know those things)

Also, remember that each individual log message needs to capture all the context, since more likely than not, there are multiple things going on in the system at once and log messages from unrelated things will be intertwined. (So ideally, the example above would have all of the entity/owner/key info reported in a _single_ message.) We need to be able to correlate log messages by looking at their contents, not just the time/order in which they were emitted.

So here's what I would suggest:

1. I know you've worked hard on this PR over the last month or so (and I'm truly sorry I didn't look at this PR/catch this earlier), but I think we should probably drop it in its current form for all the reasons I mentioned above.
2. Let's look into Winston's support for adding tags/context to logs automatically, and do a small PR to add anything that might be missing (e.g. wrapping the winston logger in our own logger class if necessary). See below.
3. Let's change all the _existing_ logs to add tags/context in a separate PR from (2).
4. Then let's look at where we need to add/update/remove logs to make the logs themselves clearer. IMO we should do this by looking at real-world log output and taking a more targeted approach than what you've done here; chances are what we have now is already "pretty good" (even if not ideal) and we can get most of the value just by focusing on areas we think are/might be problematic.

PR #677, opened 2021-04-07
Modified logic and design for update status operation

### LINKED ISSUE #665: update_status() is a race condition waiting to happen
Many, probably even most of the races I have fixed in Customer-1 code involve update_status() in some form. Usually update_status() being done against some entity other than the one being operated on, based on information that may be out of date (e.g. a simple read/modify/write).

I know there's been a lot of resistance in the past to making update_status() atomic in some way, but we really need to find some way to make it harder to misuse, because otherwise people will continue to shoot themselves in the foot. Here are a couple quick ideas. I'm sure there's more we could do, let's discuss in the comments.

### Option 1: Make it Atomic

As I'd previously proposed informally/offline, have a `status_version` field which works like `spec_version` and require atomic updates of status fields. This makes status updates more heavyweight, but ensures developers must handle update conflicts if they want their code to work. (Or at least, it will be obvious when a race would have occurred, because things will randomly fail with update conflict errors.)

This is a "brute force" approach in the sense that it covers all the use cases, but does so in a way that pushes more complexity onto the application/developer. So while I'm not ready to discard this idea out of hand, I do understand the concerns with it.

### Option 2: Restrict it to Intent Handlers

Restrict the use of update_status() to only those situations where we can be sure it is synchronized with respect to other things going on with the entity. Right now, this means construction, destruction and intent handlers. So the ONLY allowable time/place to call update_status() would be from a destructor, or a running intent handler. This means it is no longer necessary to worry about atomicity at all, since there is already higher-level synchronization being done.

This raises an interesting challenge though—there are a few places where update_status() is used that are not covered by the above restriction:
1. A status change on one entity should be reflected (possibly in a different form) on another entity
2. An "out-of-band" status change must be triggered (e.g. thru a push notification/API call against the entity, background task, etc.)

I don't quite know what to do with these yet—such things should certainly be synchronized with intent handlers, but how? 

For (1), one thought I had was to turn any such "out-of-band" updates into a spec change which then triggers the status change. For example, such entities could have a "refresh_status" or similar spec/status field which outsiders could tweak to trigger a status update. Something like this could even be baked into the engine—define a special "refresh_status" procedure which is synchronized with intent handlers, gets called periodically by the engine, and could be called by outsiders, to re-compute any status fields.

However, (2) almost demands some type of generalized synchronization, since out-of-band updates can, by definition, happen at any time and while other things are going on.

One way to address both (1) and (2) might be to expose a generalized task-scheduling mechanism (which could also be used by intent handlers) to queue tasks to run against a particular entity. The task would have access to update_status() on that entity and be synchronized with all other tasks, intent handlers, creation/deletion, etc.

I haven't spent time to refine these ideas yet, but wanted to get them out of my head and at least start the discussion more formally. What do others think?

nitesh-idnani1:
Closes #665 
- Modified return type for update spec/status in client and engine.
- Added status hash initialization using the current status value.
- Added check in status db to verify the input hash against db hash value.
- Added error type for entity status conflict (stale hash value).
- Added tests to verify error for stale hash value in update status.


papiea-engine/src/databases/status_db_mongo.ts
```diff
@@ -20,35 +21,58 @@ export class Status_DB_Mongo implements Status_DB {
...
             }, {
-                upsert: true
-            });
-        if (result.result.n !== 1) {
-            throw new PapieaException(`MongoDBError: Amount of updated entries doesn't equal to 1: ${result.result.n} for kind ${entity_ref.provider_prefix}/${entity_ref.provider_version}/${entity_ref.kind}`, { provider_prefix: entity_ref.provider_prefix, provider_version: entity_ref.provider_version, kind_name: entity_ref.kind, additional_info: { "entity_uuid": entity_ref.uuid }})
+                    $set: {
+                        "status": status,
+                    },
+                    $setOnInsert: {
+                        "metadata.status_hash": current_status_hash,
```
joshua-berry-ntnx, line 40: Why is this in `$setOnInsert` and not `$set`? Seems like we would always want to write the new status hash on a `replace_status()`.

papiea-engine/src/databases/status_db_mongo.ts
```diff
@@ -20,35 +21,58 @@ export class Status_DB_Mongo implements Status_DB {
...
+                    }
+                }, {
+                    upsert: true
+                });
+            if (result.result.n !== 1) {
+                throw new PapieaException(`MongoDBError: Amount of updated entries doesn't equal to 1: ${result.result.n} for kind ${metadata.provider_prefix}/${metadata.provider_version}/${metadata.kind}`, { provider_prefix: metadata.provider_prefix, provider_version: metadata.provider_version, kind_name: metadata.kind, additional_info: { "entity_uuid": metadata.uuid }})
+            }
+            return await this.get_status(metadata)
+        } catch (err) {
+            if (err.code === 11000) {
```
joshua-berry-ntnx, line 51: Can you add a comment (or maybe a named constant) explaining what this error code is?

papiea-engine/src/databases/status_db_mongo.ts
```diff
@@ -57,22 +81,34 @@ export class Status_DB_Mongo implements Status_DB {
...
+                    "metadata.uuid": metadata.uuid,
+                    "metadata.kind": metadata.kind
                 }, aggregrate_fields, {
                     upsert: true
                 });
-        } catch (e) {
-            if (e.code === 9) {
-                throw new PapieaException(`MongoDBError: Update body might be 'undefined', if this is expected, please use 'null' for kind ${entity_ref.provider_prefix}/${entity_ref.provider_version}/${entity_ref.kind}`,  { provider_prefix: entity_ref.provider_prefix, provider_version: entity_ref.provider_version, kind_name: entity_ref.kind, additional_info: { "entity_uuid": entity_ref.uuid }})
+        } catch (err) {
+            if (err.code === 9) {
```
joshua-berry-ntnx, line 93: Same as above -- what's this error code?

papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -589,6 +589,55 @@ describe("Provider Sdk tests", () => {
...
+        try {
+            await sdk.register();
+            const kind_name = sdk.provider.kinds[ 0 ].name;
+            const { data: { metadata, spec } } = await axios.post(`${ sdk.entity_url }/${ sdk.provider.prefix }/${ sdk.provider.version }/${ kind_name }`, {
+                spec: {
+                    x: 10,
+                    y: 11
+                }
+            });
+            await axios.post(`${ sdk.entity_url }/${ sdk.provider.prefix }/${ sdk.provider.version }/${ kind_name }/${ metadata.uuid }/procedure/moveX`, { x: 5 });
```
joshua-berry-ntnx, line 633: Do you expect this to fail? If it succeeds, your test will pass.

joshua-berry-ntnx (changes requested):

papiea-engine/src/intentful_core/entity_creation_strategies/entity_creation_strategy_interface.ts
```diff
@@ -100,6 +101,14 @@ export abstract class EntityCreationStrategy {
                 })
             request_metadata.spec_version = spec_version
         }
+        if (request_metadata.status_hash === undefined || request_metadata.status_hash === null) {
```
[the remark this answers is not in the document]
Emulebest: Agree

papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
Emulebest: Looks fine, as it has `assertions(1)` on top

papiea-sdk/typescript/__tests__/sdk_tests/provider_sdk.test.ts
```diff
@@ -591,6 +591,55 @@ describe("Provider Sdk tests", () => {
         }
     });
 
+    test("Provider with kind level procedures update status with stale status hash should fail", async () => {
```
Emulebest, line 594: Can we have a counter part to this test somewhere close that deliberately succeeds?

Emulebest (changes requested):
Overall looks good, some small comments

papiea-engine/src/utils/utils.ts
```diff
@@ -183,3 +184,7 @@ export function getVersionVerifier(enginePapieaVersion: string) {
         next();
     }
 }
+
+export function getObjectHash(obj: any) {
+    return hash(obj);
+}
```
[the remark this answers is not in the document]
nitesh-idnani1: This is an interface in object-hash, I can't import it directly here.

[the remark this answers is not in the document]
joshua-berry-ntnx: To re-export, I think you can do something like:

```ts
export {getObjectHash: hash};
```

papiea-engine/src/databases/status_db_interface.ts
```diff
@@ -1,14 +1,32 @@
...
-    // Update the status in the status db. As long as the input is
-    // correct this always succeeds.
-    replace_status(entity_ref: Provider_Entity_Reference, status: Status): Promise<void>;
+    /**
+     * Replaces status for the entity
+     * @async
+     * @method
+     * @param  {EntityStatusUpdateInput} metadata
+     * @param  {Status} status
+     * @returns Promise
```
joshua-berry-ntnx, line 15: I believe a lot of these @-directives are optional; let's leave them off unless we have some actual documentation for each of these things (e.g. what is the promised return value?)

papiea-engine/src/databases/status_db_interface.ts
```diff
@@ -1,14 +1,32 @@
...
+     */
+    replace_status(metadata: EntityStatusUpdateInput, status: Status): Promise<[Metadata, Status]>;
+
+    /**
+     * Updates status for the entity
+     * @async
+     * @method
+     * @param  {EntityStatusUpdateInput} metadata
+     * @param  {Status} status
+     * @returns Promise
```
joshua-berry-ntnx, line 26: Same as above

papiea-engine/src/databases/utils/errors.ts
```diff
@@ -1,16 +1,17 @@
...
         super({message: msg, entity_info: {provider_prefix: metadata.provider_prefix, provider_version: metadata.provider_version, kind_name: metadata.kind, additional_info: { "entity_uuid": metadata.uuid, "existing_spec_version": metadata.spec_version.toString() }}});
-        this.name = PapieaError.ConflictingEntity
-        Object.setPrototypeOf(this, ConflictingEntityError.prototype);
+        this.name = PapieaError.SpecConflictingEntity
+        Object.setPrototypeOf(this, SpecConflictingEntityError.prototype);
     }
 }
 
-export class GraveyardConflictingEntityError extends ConflictingEntityError {
+export class GraveyardConflictingEntityError extends SpecConflictingEntityError {
```
joshua-berry-ntnx, line 14: Is this only thrown for spec updates or is it thrown for status updates too? If both, we may need to either get rid of this as a separate error type or split it into SpecGraveyard and StatusGraveyard (which I'm also not too wild about since we have a weird combinatoric mix of types).

papiea-engine/src/intentful_core/intentful_strategies/status_update_strategy.ts
```diff
@@ -9,19 +9,35 @@ import { PapieaException } from "../../errors/papiea_exception"
 
 export abstract class StatusUpdateStrategy {
     statusDb: Status_DB
+    specDb: Spec_DB
```
joshua-berry-ntnx, line 12: It seems a bit weird to me that we are messing with spec at all here. I see below that we are only using it to fetch the spec to be included in the return value; is this something that could be done by the caller of the StatusUpdateStrategy?

joshua-berry-ntnx (approved):
I'm OK with this being merged as-is but there are a couple follow-up items noted below we should take on before closing the parent issue. (These are just refactoring items, not correctness or public-facing API issues.)

papiea-engine/src/intentful_core/intentful_strategies/status_update_strategy.ts
nitesh-idnani1: "DifferUpdateStrategy" class was already using the spec db as part of it's extended member variable. I just modified that as part of the base class so that the other classes can use it as well, for reading spec in this case. However, if we want to avoid the use of spec db in these classes, we can pass the spec/metadata in the function input.

papiea-engine/src/databases/utils/errors.ts
nitesh-idnani1: Yes, this error suggest conflict with the deleted entity for the user. I think the name is correct, since we want the user to understand that the conflict is with the entity and not just the spec. I don't think we can put any check/restriction for status hash since that value is not passed by user but something that the engine generates initially.

papiea-engine/src/intentful_core/intentful_strategies/status_update_strategy.ts
joshua-berry-ntnx: hm ok. I probably need to look at the code flow here a little more to understand the context. But like I mentioned, I'm OK with this for now.

PR #678, opened 2021-04-08
Pass recent entity info to destructors/intent handlers

### LINKED ISSUE #675: Papiea sometimes does not pass up-to-date/complete entity info to procedures or intent handlers
### System Details

- **Papiea Engine version**: 0.9.51
- **Papiea SDK version:** 0.9.51
- **Papiea Client version:** 0.9.51

### Describe the Bug

Sometimes it seems Papiea isn't passing fully-complete entity info to procedures, and we have to do an explicit `get()` to retrieve the full entity. For example, properties which I know are set on the entity (and which are not actively being worked on) may be missing entirely.

Most recently I've seen this happen with destructor procedures, but I think I've also seen it happen in intent handlers. Can we make sure we are passing a complete, up-to-date view of the entire entity?

### Bug information
Need to investigate further; filing as a placeholder for now. (And also filing in case someone already knows why this is happening and can provide an easy explanation/fix.)

### Expected/Actual Behavior
Every time a procedure or intent handler is invoked, I expect to receive a complete view of the entity with all the properties present (and with their most recent values at the time of invocation).

### Steps to Reproduce
I need to dig more to find out precisely when/how this happens.

### Impact
Procedures sometimes fail with errors due to the missing properties.

### Additional Details

<hr>

nitesh-idnani1:
Closes #675
- Fixed deep copy bug in differ sanitizer functions
- Modified destructor handler input to include the complete entity structure.

### DIFF
--- a/papiea-engine/src/intentful_core/sfs_compiler.ts
+++ b/papiea-engine/src/intentful_core/sfs_compiler.ts
@@ -2,6 +2,7 @@ import { ValidationError } from "../errors/validation_error"
 import {DiffContent, Spec, Status} from "papiea-core"
 import { PapieaException } from "../errors/papiea_exception";
 import { Logger } from "papiea-backend-utils";
+import { cloneDeep } from "lodash"
 
 // TODO: add d.ts for type annotations
 const papi_clj = require("../../papiea-lib-clj/papiea-lib-clj.js").papiea_lib_clj;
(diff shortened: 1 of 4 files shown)


papiea-engine/src/intentful_core/sfs_compiler.ts
```diff
@@ -2,6 +2,7 @@ import { ValidationError } from "../errors/validation_error"
 import {DiffContent, Spec, Status} from "papiea-core"
 import { PapieaException } from "../errors/papiea_exception";
 import { Logger } from "papiea-backend-utils";
+import { cloneDeep } from "lodash"
```
Emulebest, line 5: Please remove lodash. This is a bad idea to use it for a single function. `JSON.parse(JSON.stringify(entity))` does the job

Emulebest (changes requested):
Please add a test and remove lodash, thanks!

PR #681, opened 2021-04-16
Added support for HTTPS endpoints in papiea engine

### LINKED ISSUE #488: Papiea endpoints should be HTTPS in production

nitesh-idnani1:
Closes #488
- Modified papiea client to send https requests to the engine.
- Modified all tests to use https endpoints.
- Modified express logic to create a HTTPS server on same port config.
- Modified python and typescript SDKs to send https requests.


nitesh-idnani1:
Creating a draft PR, since few things need to be resolved before dev/master test runs.

- Auth tests are failing due to these changes, need some help debugging those.
- Need to deploy certificates for server & client on CircleCI environment.

Emulebest:
@nitesh-idnani1 what is the status of this PR? Have you solved the auth test problems?

nitesh-idnani1:
> @nitesh-idnani1 what is the status of this PR? Have you solved the auth test problems?

Yes, I've solved that. You can start reviewing this one

papiea-client/src/entity_client.ts
```diff
@@ -242,20 +252,28 @@ export interface EntityCRUD {
     close(): void
 }
 
-export function kind_client(papiea_url: string, provider: string, kind: string, version: string, s2skey?: string, tracer?: Tracer): EntityCRUD {
+export function kind_client(papiea_url: string, provider: string, kind: string, version: string, s2skey?: string, ca_path?: string, key_path?: string, cert_path?: string, tracer?: Tracer): EntityCRUD {
```
Emulebest, line 255: Can we move `httpsAgent` to be save here once instead of passing it to all the functions?

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
```diff
@@ -127,7 +128,10 @@ export class ProviderSdk implements ProviderImpl {
             headers: {
                 'Content-Type': 'application/json',
                 'Authorization': `Bearer ${this._s2skey}`
-            }
+            },
+            httpsAgent: new https.Agent({  
+                rejectUnauthorized: false
```
Emulebest, line 133: Are these tweakable? In production we should be able to set these to `true`

Emulebest (changes requested):

papiea-client/src/entity_client.ts
nitesh-idnani1: We need to pass it with the headers while making axios requests, can't do that here.

papiea-sdk/typescript/src/provider_sdk/typescript_sdk.ts
nitesh-idnani1: We can use NODE_ENV variable to configure this, what do you suggest?

papiea-engine/src/auth/oauth2.ts
```diff
@@ -32,6 +34,13 @@ function convertToSimpleOauth2(description: any) {
             authorizePath: oauth.authorize_uri,
             revokePath: oauth.revoke_uri
         },
+        http: {
+            agents: {
+                https: new https.Agent({ rejectUnauthorized: false }),
+                http: new http.Agent(),
+                httpsAllowUnauthorized: new https.Agent({ rejectUnauthorized: false })
+            }
+        },
```
joshua-berry-ntnx, line 43: Not sure I understand what this is doing / why it's needed. Can you explain? Why does oauth need its own HTTP agent?

Also, both `https` and `httpsAllowUnauthorized` seem to have the same settings.

papiea-client/src/entity_client.ts
joshua-berry-ntnx: Same as above; let's pass the agent itself instead of the certs

papiea-client/src/entity_client.ts
```diff
@@ -242,20 +252,28 @@ export interface EntityCRUD {
...
     const client_tracer = tracer ?? ClientTracer.getTracer()
     const close_func = tracer ? (tracer as any).close() : ClientTracer.close
     const the_s2skey = s2skey ?? 'anonymous'
+    let httpsAgent = new https.Agent({
+        rejectUnauthorized: false
```
joshua-berry-ntnx, line 260: Breaks HTTPS authentication as mentioned above

joshua-berry-ntnx (changes requested):
In general we should never do `rejectUnauthorized: false` except in specific testing situations (and even then, IMO we shouldn't have to do it). Instead, I think the right thing to do is pass a proper CA (i.e. `ca.crt`) instead of setting `rejectUnauthorized`.

My concern about doing this even in tests is that someone else will later come along and see this, not realize its significance, and think it's a correct thing to do in shipping code (it very much is not). If we are "secure by default" even in our test/example code, hopefully we can save someone else some headache down the road.

I think most/all of my comments are about this one issue; also had a question about the oauth changes you made, see below.

papiea-engine/src/auth/oauth2.ts
nitesh-idnani1: Simple OAuth2 requires all three parameters, if we want to set a https agent.

Emulebest (approved):

papiea-engine/install_certificates.sh
```diff
@@ -0,0 +1,17 @@
...
+openssl req -new -key server.key -out server.csr -subj "/C=US/ST=California/L=SanJose/O=Nutanix/OU=Papiea/CN=localhost" &&
+openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 365 &&
+cd ../../papiea-client &&
+mkdir -p certs &&
+cd certs &&
+cp ../../papiea-engine/certs/ca.crt ./ &&
+cd ../../papiea-sdk/python/e2e_tests &&
+cp -r ../../../papiea-client/certs ./ &&
+cd ../../typescript &&
+cp -r ../python/e2e_tests/certs ./
```
joshua-berry-ntnx, line 17: Could we just symlink the entire `certs` dir wherever it's needed? Might be a little simpler than copying the individual files around

papiea-engine/src/main.ts
```diff
@@ -99,7 +103,8 @@ async function setUpApplication(): Promise<express.Express> {
 }
 
 setUpApplication().then(app => {
-    app.listen(serverPort, function () {
-        console.info(`Papiea app listening on port ${serverPort}!`);
+    const https = require('https');
+    https.createServer({ key: privateKey, cert: certificate, request_cert: true }, app).listen(serverPort, function () {
```
joshua-berry-ntnx, line 107: I don't think we need to `request_cert: true` since we're not planning to authenticate the client.

papiea-engine/start.sh
```diff
@@ -6,6 +6,7 @@ if [ -d "/node_modules" ] && [ ! -d "node_modules" ]; then
 fi
 
 npm run build-clj
+bash install_certificates.sh
```
joshua-berry-ntnx, line 9: If we're going to unconditionally generate certs, we should be checking at each step whether we need to generate that particular file, otherwise we will overwrite whatever the user has provided.

papiea-engine/install_certificates.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -e
+mkdir -p certs &&
+cd certs &&
+openssl genrsa -out ca.key 4096 &&
+openssl req -x509 -new -nodes -key ca.key -days 365 -out ca.crt -subj "/C=US/ST=California/L=SanJose/O=Nutanix/OU=CA/CN=papiea" &&
+openssl genrsa -out server.key 4096 &&
+openssl req -new -key server.key -out server.csr -subj "/C=US/ST=California/L=SanJose/O=Nutanix/OU=Papiea/CN=localhost" &&
+openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 365 &&
```
joshua-berry-ntnx, line 9: As noted below, for each of these, we should be checking whether it already exists first.

joshua-berry-ntnx (changes requested):
Looks mostly good apart from `install_certificates.sh`, where it seems we're unconditionally overwriting whatever certs are there (which is probably not what we want).

papiea-engine/install_certificates.sh
```diff
@@ -1,17 +1,33 @@
...
+mkdir -p certs
+cd certs
+
+if [ ! -f ca.crt ]; then
+    # generate ca key and certificate
+    openssl genrsa -out ca.key 4096
+    openssl req -x509 -new -nodes -key ca.key -days 365 -out ca.crt -subj "/C=US/ST=California/L=SanJose/O=Nutanix/OU=CA/CN=papiea"
+fi
+
+if [ ! -f server.crt ]; then
```
joshua-berry-ntnx, line 14: Also check for server.key

joshua-berry-ntnx (approved):
LGTM. Since this is a breaking change, let's notify Customer-1 when it's merged.

Issue #679, opened 2021-04-16
Replace all list_*() methods in database interfaces with async iterators


joshua-berry-ntnx:
Virtually every database interface in Papiea seems to have some sort of list() method which fetches all matching entities and returns an array. This will not scale beyond small database sizes because we will run out of memory very quickly.

All of these APIs need to be replaced with APIs that return async iterators, so we never load the entire result set into memory at once.

joshua-berry-ntnx:
BTW, this has already caused a severe out-of-memory situation in Papiea right now, where if we have too many intent watchers, we cannot process them because there are places where we try to load the entire list into memory.

Worse, this is done at startup in the diff_resolver process, so the diff_resolver process almost immediately crashes and we cannot make progress on resolving any diffs. So mitigating this is a blocker issue right now (although a full fix can come later).

Emulebest:
Let's discuss the possible alternatives here for concrete use cases

PR #683, opened 2021-04-19
Fix OOM issue && Use AsyncIterators to fetch intent watchers

joshua-berry-ntnx:
If we have accumulated a lot of intent watchers in the DB due to spec changes,
they may not all fit in memory. Indeed, we have seen OOM issues in production
because of this.

We also saw another OOM issue due to a complete lack of flow control in the
intent_resolver—if we have too many intent watchers to update, the process of
updating them all may take so long that we start the next round, and
progressively run more and more updates in parallel until we run out of memory.

Makes some progress on #679.

### DIFF
--- a/papiea-engine/Dockerfile
+++ b/papiea-engine/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get update && apt-get upgrade -y \
     && apt-get install -y wget curl openjdk-11-jdk \
     # For nodejs
     && apt-get install -y gnupg \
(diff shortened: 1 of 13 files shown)


nitesh-idnani1:
The PR looks good, will finish my review once the tests are stable.

joshua-berry-ntnx (commented):
Your changes to my changes look good. :) Will go ahead and merge, thanks for your review and help!

PR #706, opened 2021-05-25
Added flag to disable https in papiea engine

### LINKED ISSUE #705: Add explicit option to turn off HTTPS
For backward compatibility for the time being, we want to be able to explicitly turn off HTTPS via an environment variable. The default should still be HTTPS-enabled, though.

nitesh-idnani1:
Closes #705

### DIFF
--- a/papiea-engine/src/main.ts
+++ b/papiea-engine/src/main.ts
@@ -48,7 +48,7 @@ const prettyPrint = config.pretty_print
 const tracingConfig = config.tracing_config
 const privateKey = readFileSync(resolve(__dirname, config.server_key_path), "utf-8")
 const certificate = readFileSync(resolve(__dirname, config.server_cert_path), "utf-8")
-const disableHTTPS = config.disable_https
+const https = config.https
 
 async function setUpApplication(): Promise<express.Express> {
     const logger = LoggerFactory.makeLogger({level: loggingLevel, verbosity_options: verbosityOptions, pretty_print: prettyPrint});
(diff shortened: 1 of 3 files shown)


papiea-engine/src/main.ts
```diff
@@ -48,6 +48,7 @@ const prettyPrint = config.pretty_print
 const tracingConfig = config.tracing_config
 const privateKey = readFileSync(resolve(__dirname, config.server_key_path), "utf-8")
 const certificate = readFileSync(resolve(__dirname, config.server_cert_path), "utf-8")
+const disableHTTPS = config.disable_https
```
Emulebest, line 51: Can we make this variable `config.https` and make the default value of it as `true`? So that we can tweak it with `PAPIEA_HTTPS=false` instead of `PAPIEA_DISABLE_HTTPS=true`

Emulebest (changes requested):
Does this PR also affect SDK and test if the flag is set?

papiea-engine/src/main.ts
joshua-berry-ntnx: Agree, this is a very good point.

joshua-berry-ntnx:
> Does this PR also affect SDK and test if the flag is set?

Agree, just for simplicity we should probably cover the SDK. (We may need to make this more fine-grained later but for now it's sufficient to have the same flag apply to both engine and SDK—so the whole stack is either HTTPS or it's not.)

nitesh-idnani1:
> > Does this PR also affect SDK and test if the flag is set?
> 
> Agree, just for simplicity we should probably cover the SDK. (We may need to make this more fine-grained later but for now it's sufficient to have the same flag apply to both engine and SDK—so the whole stack is either HTTPS or it's not.)

There's no need for us to include HTTPS config specification in the sdk. All the config params are controlled by the user during provider setup so they can control it to be HTTP or HTTPS.
The use of flag can be done to validate the requests (HTTP or HTTPS), otherwise I don't think we need it.

joshua-berry-ntnx (approved):

PR #713, opened 2021-07-23
Log objects before message for pino

manojsudheendra:


nitesh-idnani1 (approved):
LGTM. Can you also add the timestamp change as part of this PR (it's just a one-liner)?

manojsudheendra:
> LGTM. Can you also add the timestamp change as part of this PR (it's just a one-liner)?

I tried that. It was still showing logs with the epoch time. I found out that there is a tool called pino-pretty which we can use for post processing which can change the timestamp display.

